PR #27, opened 2022-04-18
Extract game info from pgn

nocommentcode:
- Created chessEngine.ts, our wrapper for our chess engine library
- Implemented the parsePgn method of wrapper
- Called parsepgn from arbiterSSetupMode after fetching the pgn from the api
- Wrote tests to test the new chessEngine file


gestureleft (approved):
💯

PR #31, opened 2022-04-19
Implement the result type

gestureleft:
This PR introduces a convenient way of representing computations that could fail. Anytime a function could 'fail', it should return a `Result`. Consider a function that could either return a `number` successfully, or fail abysmally. With this new mechanism, it might look like this:
```ts
const mightFail = (someParam: number): Result<number> => {
  if (someParam === 'succeed please') {
    return succ(180);
  }
  return fail('boom');
}
```

This PR also introduces a global React context for storing an error string. It gets rendered as a toast, complete with nice animation and icon:

https://user-images.githubusercontent.com/57308619/164176332-7397fdd2-30ba-433f-bcff-85634bff9181.mov


gwenllian1 (changes requested):
just need to fix the issue you pointed out where it is only checking arbiter setup mode :)

gwenllian1 (approved):

PR #32, opened 2022-04-20
Updated dependencies

gestureleft:
There was a vulnerability in the version of cocoapods-downloader that we
were using, so let's bump the version.

May as well bump other package versions while we're here


B-Gunawan (approved):
Not a super major semver updates, looks good to me.

PR #34, opened 2022-04-21
Implement the App's Toolbar

gestureleft:
This PR introduces three new components:
- `Toolbar`
- `IconButton`
- `Sheet`

The Toolbar component is the App's toolbar, and will allow the user to fetch help and make various other actions based on the current App Mode. The colour of the toolbar responds to the mode the user is in - as well as the colour of the text. On dark colours, the text is white, and on light colours, the text is black (see below screenshot for example)

The IconButton component is a "pressable" icon, that allows us to have icons that function as buttons.

The Sheet component is a generic "modal" dialog, that supports rendering a title and come content. This will be helpful for rendering help text to the user.

https://user-images.githubusercontent.com/57308619/164387771-ab478058-e8b7-4e4c-b1f3-1722bb0e5997.mov

Toolbar Responsive Text Color Demo


packages/TorneloScoresheet/src/components/IconButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const IconButton: React.FC<IconButtonProps> = ({
+  icon,
+  onPress,
+  colour,
+  style,
+}) => {
+  return (
+    <Pressable onPress={onPress}>
+      <Icon style={style} size={40} color={colour} name={icon} />
```
nocommentcode, line 26: Do we want the size to be constant at 40?

packages/TorneloScoresheet/src/components/Toolbar.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const currentColour = colourForMode[appModeState.mode];
+  const currentTextColour = textColour(currentColour);
+  return (
+    <>
+      <StatusBar barStyle={statusBarStyleForColor(currentColour)} />
+      <Sheet
+        title="Help"
+        dismiss={() => setShowSheet(false)}
+        visible={showSheet}
+        content="Put help here..."
```
nocommentcode, line 51: is the sheet component only going to be for instructions on how to use each mode in the app? If so how do we plan to set the content? Should we have a Record<appMode, string> like you did for colour or are we planning on using the sheet for other stuff too?

nocommentcode (commented):
cool tings

packages/TorneloScoresheet/src/components/Toolbar.tsx
gestureleft: I think the idea at this stage is that we'll have a separate file that has a Record of all of the help strings, like you suggest. This concept shouldn't be tied to the `Sheet` component though, since we may want to use sheets for other things.
So we'll have a file with help strings that the `Toolbar` will import and pass the correct string to the `Sheet`

PR #36, opened 2022-04-27
Table pairing selection table

nocommentcode:
Arbiter mode major work.
Arbiter mode now has 2 views:

1. Enter Pgn view
2. Table pairing selection view

- Updated the styling of the enter pgn view
image

- We now have the table pairing selection view where the arbiter can select which paring is on this table
image

- The arbiter can select a pairing and then confirm
image

The styling definitely still needs some work, we are waiting for the final design from the client but the functionality is here


.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,10 @@
+{
+  "ExpandedNodes": [
+    "",
+    "\\packages",
+    "\\packages\\TorneloScoresheet",
+    "\\packages\\TorneloScoresheet\\src"
+  ],
+  "SelectedNode": "\\packages\\TorneloScoresheet\\src\\arbiterSetupMode.ts",
+  "PreviewInSolutionExplorer": false
+}
```
gestureleft, line 10: Don't know what any of this is - but I don't think the `.vs` directory is meant to be in this PR? I think this is some editor config or smt that only applies to your particular editor setup?

packages/TorneloScoresheet/src/pages/TablePairingSelection.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  return (
+  <View>
+      <Text onPress={returnToPgnLinkPage} style={styles.backBtn}>{'<'} Back</Text>
+      <Text style={styles.instructionTitle}>Boards</Text>
+      <Text style={styles.instructionContents}>Select the board that this iPad is assigned to. This can be changed later.</Text>
+      <FlatList scrollEnabled={true} style={styles.pairingList} data={appModeState.pairings} renderItem={renderPairing}>
+
+      </FlatList>
+
```
gestureleft, line 93: Let's make this self-closing (as suggested by eslint). Something like:

```js
<FlatList
  scrollEnabled={true}
  style={styles.pairingList}
  data={appModeState.pairings}
  renderItem={renderPairing} />
```

packages/TorneloScoresheet/src/pages/Main.tsx
```diff
@@ -3,19 +3,20 @@ import { Text } from 'react-native';
...
-  );
+  switch(appMode){
+    case AppMode.ArbiterSetup:
+      return <ArbiterSetup />
+    case AppMode.TablePairing:
+      return <TablePairingSelection />
+    default:
+      return <Text>Unsupported app mode</Text>
+
+  }
```
gestureleft, line 19: This file needs to be formatted

packages/TorneloScoresheet/src/context/AppModeStateContext.tsx
```diff
@@ -1,36 +1,42 @@
 import React, { useContext, useState } from 'react';
-import { makeEnterTablePairingMode } from '../arbiterSetupMode';
-import { AppModeState, AppMode } from '../types/AppModeState';
+import { makegoToTablePairingSelection } from '../goToTablePairingSelection';
```
gestureleft, line 2: Camel case `makegoTo..` -> `makeGoTo`

packages/TorneloScoresheet/src/pages/ArbiterSetup.tsx
```diff
@@ -2,48 +2,23 @@ import React, { useState } from 'react';
...
-    <View style={styles.arbiterSetup}>
-      <View>
-        <Text>Enter Game Link:</Text>
-        <TextInput onChangeText={setUrl} value={url} placeholder="Game Link" />
-      </View>
-      <Button title="Next" onPress={handleNextClick} />
-    </View>
-  );
+  const [appModeState] = useAppModeState();
+  if (appModeState.mode != AppMode.ArbiterSetup) {
```
gestureleft, line 9: This should be `!==` (as per eslint)

packages/TorneloScoresheet/src/pages/EnterPgnLink.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { useError } from '../context/ErrorContext';
+import { AppMode, ArbiterModeViews } from '../types/AppModeState';
+import { isError } from '../types/Result';
+
+const EnterPgnLink: React.FC = () => {
+  const [appModeState, { goToTablePairingSelection }] = useAppModeState();
+  const [url, setUrl] = useState('');
+  const [, showError] = useError();
+
+  if (appModeState.mode != AppMode.ArbiterSetup) {
```
gestureleft, line 16: This should be `!==` (as per eslint)

packages/TorneloScoresheet/src/pages/TablePairingSelection.tsx
```diff
@@ -1,106 +1,174 @@
...
   return (
-  <View>
-      <Text onPress={returnToPgnLinkPage} style={styles.backBtn}>{'<'} Back</Text>
-      <Text style={styles.instructionTitle}>Boards</Text>
-      <Text style={styles.instructionContents}>Select the board that this iPad is assigned to. This can be changed later.</Text>
-      <FlatList scrollEnabled={true} style={styles.pairingList} data={appModeState.pairings} renderItem={renderPairing}>
+    <View>
+      <View style={styles.buttonContainer}>
+        <Text onPress={goToEnterPgnLink} style={styles.backBtn}>
+          {'<'} Back
```
gestureleft, line 89: Let's use a chevron from the "react-native-vector-icons" we have installed here? Don't necessarily have to this in this PR, since this PR isn't a final iteration of the UI, just wanted to flag it

packages/TorneloScoresheet/src/pages/TablePairingSelection.tsx
```diff
@@ -1,106 +1,174 @@
...
+        </Text>
+        {renderConfirm()}
```
gestureleft, line 91: Having a separate function for a specific part of a component is a bit unusual.

Usually I would write this as:
```js
...
</Text>
{showConfirmButton && selectedPairing !== null &&
  <Text  onPress={() => goToTablePairingMode(selectedPairing)} style={styles.forwardBtn}>
      Confirm {'>'}
  </Text>}
...
```

The advantage is that all of the 'presentation' logic is all at the bottom of the function in the return statement, and all of the 'business' logic is in the hooks, callbacks, etc. above. It is also more readable and concise imo.

gestureleft: But also happy to hear what your thoughts are on this - this is just how *I* would do it, so not necessarily the best way...

packages/TorneloScoresheet/src/types/AppModeState.ts
```diff
@@ -7,14 +7,20 @@ export enum AppMode {
...
 
+export enum ArbiterModeViews {
+  EnterPgnLink,
+  TablePairingSelection
+}
 export type ArbiterSetupMode = {
   mode: AppMode.ArbiterSetup;
+  view: ArbiterModeViews
+  games?: number;
+  pairings?: GameInfo[];
```
gestureleft, line 18: I'm thinking: rather than having this one mode have two views, why don't we just invent a new mode that happens before arbiter setup mode? Like PGNDownload mode or something (we can come up with a better mode). I think this would make the state a bit more easy to understand? This also works though - just a though, let me know what you think

packages/TorneloScoresheet/__tests__/useAppModeState.test.ts
```diff
@@ -32,4 +32,53 @@ describe('useAppModeState', () => {
...
+      wrapper: AppModeStateContextProvider,
+    });
+
+    // go into table pairing selection view
+    await act(async () => {
+      await result.current[1].goToTablePairingSelection(pgnUrl);
+    });
+
+    // go back to pgn enter view
+    result.current[1].goToEnterPgnLink();
```
gestureleft, line 76: I think this should be in a call to `act`? This is causing the test to fail I think

gestureleft (changes requested):
Awesome work! I left some feedback that I'd like to hear your thoughts on. Also, please run the formatter on all of the files you've touch - some of them aren't formatted :))

.vs/VSWorkspaceState.json
gestureleft: So I think `.vs` should be in our `.gitignore`?

packages/TorneloScoresheet/src/pages/TablePairingSelection.tsx
nocommentcode: yeh just not bothering since the desing might change, well change it later

packages/TorneloScoresheet/src/pages/TablePairingSelection.tsx
nocommentcode: yep all good just couldn't figure out how to have an if section in the middle of elements

packages/TorneloScoresheet/src/types/AppModeState.ts
nocommentcode: Yeh will make a lot more sense to have 1 view per state, i changed it now, how should we handle the state transition fuctions though, its gonna get pretty messy once we have a lot of transitions from different views, (idk about my naming convention too, the function names are a little bit long)

gestureleft: Hmm yeah. I think long names are fine tbh - probably more important to make it clear what the function is doing then save characters imo. I also *think* the state transition stuff will be fine since it will all be defined in code that is completely decoupled from the views. It might be worth considering moving some of this stuff out of the global state and into local state tbh. Let's discuss this further in the sprint retro and come up with a strategy

gestureleft (approved):
Good stuff! Great changes

PR #37, opened 2022-04-27
change chess engine to support illigal moves + tests

nocommentcode:
Support illegal moves using chess engine via a new 'force_move' function.
Instead of algebraic notation, this function will take a partial move object (with to and from properties) and will force the move to occur unless:

- There is no peice in the from square
- the peice in the from square does not belong to the player
- the peice is moved off the board
- there is a peice on the to square that belongs to the current player

apart from those conditions, the move will be processed. This fucntion does not break the en passant checks, castling or anything else.

### DIFF
--- a/.github/workflows/ios_pipeline.yml
+++ b/.github/workflows/ios_pipeline.yml
@@ -6,13 +6,12 @@
 name: Node.js CI
 
 on:
-  push:
-    branches: [main]
-  # pull_request:
-  #   branches: [ main ]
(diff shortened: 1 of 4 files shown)


nocommentcode (commented):
testing pipeline running

gestureleft (approved):
Looks good to me! Nice work 🔥

gestureleft:
Although it will need rebasing with main now that it's behind

PR #41, opened 2022-04-29
refractored state handling solution

nocommentcode:
As discussed in the meeting yesterday,
Change our handling of the state, we now have:

- Global state ( only has app mode, only used by the main view)
- EnterPgnState (uses the appmode context)
- PairingSelectionState (uses the appmode context)

### DIFF
--- a/packages/TorneloScoresheet/src/context/AppModeStateContext.tsx
+++ b/packages/TorneloScoresheet/src/context/AppModeStateContext.tsx
@@ -9,12 +9,10 @@ const AppModeStateContext = React.createContext<
   [AppModeState, React.Dispatch<React.SetStateAction<AppModeState>>]
 >([{ mode: AppMode.EnterPgn }, () => undefined]);
 
-type AppModeStateHookType = [AppModeState];
-
-export const useAppModeState = (): AppModeStateHookType => {
+export const useAppModeState = (): AppModeState => {
   const [appModeState] = useContext(AppModeStateContext);
 
-  return [appModeState];
+  return appModeState;
 };
 
 export const AppModeStateContextProvider: React.FC = ({ children }) => {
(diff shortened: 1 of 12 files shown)


packages/TorneloScoresheet/src/context/AppModeStateContext.tsx
```diff
@@ -1,48 +1,17 @@
...
-import { makegoToTablePairingSelection as makeEnterPgnToPairingSelection } from '../goToTablePairingSelection';
 import { AppModeState, AppMode } from '../types/AppModeState';
-import { GameInfo } from '../types/chessGameInfo';
-import { Result } from '../types/Result';
 
 // The global state for the app
-const AppModeStateContext = React.createContext<
+export const AppModeStateContext = React.createContext<
   [AppModeState, React.Dispatch<React.SetStateAction<AppModeState>>]
 >([{ mode: AppMode.EnterPgn }, () => undefined]);
```
gestureleft, line 8: I think we should either not export this, or document it saying that it should not be imported out side of this directory. Otherwise, down the track (even after we're done maintaining the project), somebody might import this to a random view and mutate the state directly.

This is why we hade the `makeEnterTablePgnToPairingSelection` function that takes the setter as an argument - it means that we can give that specific function the ability to mutate the global state without exporting it globally.

Ideally we'd have all of the transition functions in this function so that there is no way to leak the `setAppModeState` function, but I think that passing it as a parameter, rather than exporting it is the best compromise.

There might be a better way though, let me know what you think

packages/TorneloScoresheet/src/context/AppModeStateContext.tsx
```diff
@@ -1,48 +1,17 @@
...
-  AppModeState,
-  {
-    enterPgnToPairingSelection: (
-      liveLinkUrl: string,
-    ) => Promise<Result<undefined>>;
-    pairingSelectionToTablePairing: (pairing: GameInfo) => void;
-    pairingSelectionToEnterPgn: () => void;
-  },
-];
+type AppModeStateHookType = [AppModeState];
```
gestureleft, line 9: No need to have this type, or make it a tuple anymore I don't think. We can just have the hook return `AppModeState` directly

gestureleft (changes requested):
Looks good! I'm a bit nervous about exporting the global state context - see my comment and let me know what you think

gestureleft (approved):
LGTM

PR #42, opened 2022-04-30
System design diagrams

B-Gunawan:


documentation/DesignDocumentation.md
```diff
@@ -6,24 +6,48 @@ The two main forms of design that are most relevant to the Tornelo Scoresheet ap
...
+![architecture_diagram](Pictures/system_design-dependency_structure.png)
+
+The project's architecture is broken down to 6 parts:
+1. Pages
+    * Serves a distinct user function, as described in section 2 - Wireframes
+    * Contains components
+2. Components
+    * Parts of a page
+    * Can contain other components
+3. Context
```
nocommentcode, line 31: We have changed the way we are handling state, so I think we should include here the fact that each page will use it's own custom hook, and these hooks each access the context, but each page does not load the context itself

documentation/DesignDocumentation.md
```diff
@@ -6,24 +6,48 @@ The two main forms of design that are most relevant to the Tornelo Scoresheet ap
...
+    * Allows information to be transferred from one page to another without dependency pollution.
+    * e.g. if a component dependency chain looks like this: `A->B->C->D->E`, `A` and `E` can share data without going through `B,C,D`
+4. Types
+    * Local Typescript types, interfaces, templates and the like.
+    * Avoids cluttering other packages.
+5. Utils
+    * General utilities such as parsing, string formatters, etc.
+6. Styles
+    * Stylesheet files
+7. External packages
```
nocommentcode, line 41: we also have a hooks directory now, where the custom hook for each page is declared

nocommentcode (changes requested):
Looks good, we just need to update the way the state is handled, you might want to wait for my pr to be merged, have a look in the refractor_state branch, we have a global context that is used by multiple different hooks which access the context themselves

nocommentcode (approved):
Well done, looks great its super detailed

PR #56, opened 2022-05-06
Chess board component

nocommentcode:
Created  `ChessBoard` and `ChessSquare`  components that will be used to render the chess board.
It looks like this:
board

### DIFF
--- a/packages/TorneloScoresheet/src/components/ChessBoard/ChessBoard.tsx
+++ b/packages/TorneloScoresheet/src/components/ChessBoard/ChessBoard.tsx
@@ -1,60 +1,32 @@
 import React from 'react';
-import { Text, View } from 'react-native';
-import { colours } from '../../style/colour';
-import { Piece } from '../../types/ChessMove';
-import ChessPiece from '../ChessPiece/ChessPiece';
+import { View } from 'react-native';
+import { ChessBoardPositions } from '../../types/ChessBoardPositions';
+import ChessSquare from '../ChessSquare/ChessSquare';
+import RoundedView from '../RoundedView/RoundedView';
 import { styles } from './style';
 
 export type ChessBoardProps = {
-  board: (Piece | null)[][];
+  boardPositions: ChessBoardPositions;
   flipBoard: boolean;
 };
 
-const ChessBoard: React.FC<ChessBoardProps> = ({ board, flipBoard }) => {
-  if (flipBoard) {
-    board = board.reverse();
-  }
-  const squareColor = (row: number, col: number) => {
-    if (row % 2 === 0) {
-      if (col % 2 == 0) {
-        // even, even
-        return {
-          backgroundColor: colours.primary,
-        };
-      }
-
-      //even, odd
-      return {
-        backgroundColor: colours.secondary,
-      };
-    } else {
-      if (col % 2 == 0) {
-        // odd, even
-        return {
-          backgroundColor: colours.secondary,
-        };
-      }
-
-      // odd, odd
-      return {
-        backgroundColor: colours.primary,
-      };
-    }
-  };
-
+const ChessBoard: React.FC<ChessBoardProps> = ({
+  boardPositions,
+  flipBoard,
+}) => {
   return (
-    <View style={styles.board}>
-      {board.map((row, rowIdx) => (
-        <View style={styles.boardRow} key={'row-' + rowIdx.toString()}>
-          {row.map((peice, colIdx) => (
-            <View
-              style={[styles.boardSquare, squareColor(rowIdx, colIdx)]}
-              key={'square-' + rowIdx.toString() + colIdx.toString()}>
-              {peice && <ChessPiece piece={peice} />}
+    <View>
+      <RoundedView style={styles.board}>
+        {(flipBoard ? [...boardPositions].reverse() : boardPositions).map(
+          (row, rowIdx) => (
+            <View style={styles.boardRow} key={'row-' + rowIdx.toString()}>
+              {row.map(square => (
+                <ChessSquare square={square} key={square.position} />
+              ))}
             </View>
-          ))}
-        </View>
-      ))}
+          ),
+        )}
+      </RoundedView>
     </View>
   );
 };
--- a/packages/TorneloScoresheet/src/components/ChessBoard/ChessBoard.tsx
+++ b/packages/TorneloScoresheet/src/components/ChessBoard/ChessBoard.tsx
@@ -1,0 +1,62 @@
+import React from 'react';
+import { Text, View } from 'react-native';
+import { colours } from '../../style/colour';
+import { Piece } from '../../types/ChessMove';
+import ChessPiece from '../ChessPiece/ChessPiece';
+import { styles } from './style';
+
+export type ChessBoardProps = {
+  board: (Piece | null)[][];
+  flipBoard: boolean;
+};
+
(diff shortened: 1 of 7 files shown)


packages/TorneloScoresheet/src/components/ChessSquare/ChessSquare.tsx
```diff
@@ -0,0 +1,56 @@
...
+      };
+    }
+
+    // odd, odd
+    return {
+      backgroundColor: colours.darkBlue,
+    };
+  }
+};
+
```
B-Gunawan, line 45: This function can be simplified.

| col index | row index | sum of indexes | colour |
| --------- | ---------- | ---- | ------ |
| even | even | even | darkBlue |
| even | odd | odd | lightBlue |
| odd | even | even | darkBlue |
| odd | odd | odd | lightBlue |

Colour depends on `(row + col) % 2`

```TypeScript
const squareColor = (position: Position) => {
  const [row, col] = boardPositionToIdx(position);
  const isAlternateTile = (row + col) % 2 === 0
  return { backgroundColor: isAlternateTile ? colours.lightBlue : colours.darkBlue }
}
```

packages/TorneloScoresheet/src/types/ChessMove.ts
```diff
@@ -29,3 +34,65 @@ export type ChessMove = {
...
+export const boardIndexToPosition = (
+  rowIdx: number,
+  colIdx: number,
+): Position => {
+  const row = rowIdxToRow[rowIdx];
+  const col = colIdxToCol[colIdx];
+  return `${col}${row}`;
+};
+
+export const boardPositionToIdx = (position: Position): [number, number] => {
```
B-Gunawan, line 89: `boardPositionToIndex` may be a better function name here, just to keep consistency with `boardIndexToPosition`

packages/TorneloScoresheet/src/components/ChessBoard/ChessBoard.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { View } from 'react-native';
+import { BoardPosition } from '../../types/ChessMove';
+import ChessSquare from '../ChessSquare/ChessSquare';
+import RoundedView from '../RoundedView/RoundedView';
+import { styles } from './style';
+
+export type ChessBoardProps = {
+  board: BoardPosition[][];
```
B-Gunawan, line 9: Replace prop `board` to `boardPositions`.

packages/TorneloScoresheet/src/style/colour.ts
```diff
@@ -50,6 +55,8 @@ export const textColour = (colour: ColourType): string => {
     case colours.negative:
     case colours.darkenedElements:
     case colours.black:
+    case colours.lightBlue:
+    case colours.darkBlue:
```
B-Gunawan, line 59: Out of scope for this PR, but we may need to change the names here from `colours` to `colors` to keep consistency with CSS and and most other functions.

B-Gunawan (changes requested):

packages/TorneloScoresheet/src/style/colour.ts
nocommentcode: yeh maybe well have to have  a chat about it

gestureleft: Yeah the approach so far has been to consistently use `colour` for anything we write. Obviously we can't control the terminology React Native uses (`color`). I think it's fine that we use `colour`, because it's immediately obvious that it's our code, as long as we are consistent. I don't really care either way tho

gestureleft:
Also - no biggy if you can't be bothered - but I recommend doing an interactive rebase to make it as if the PR Review recommendations were implemented in the original commits; no need to have a random commit for fixing feedback when you can make it as if you went with that approach originally. Let me know if u want me to show you how or if u have strong opinions on this. Also let me know if I'm being too pedantic about git 😉

packages/TorneloScoresheet/src/components/ChessBoard/ChessBoard.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { View } from 'react-native';
+import { BoardPosition } from '../../types/ChessMove';
+import ChessSquare from '../ChessSquare/ChessSquare';
+import RoundedView from '../RoundedView/RoundedView';
+import { styles } from './style';
+
+export type ChessBoardProps = {
+  boardPositions: BoardPosition[][];
```
gestureleft, line 9: I think this will always be 8-by-8 yeah? We might want to make this a tuple of tuples. Always prefer the stricter type definition - in this case, the type doesn't specify how many `BoardPosition`s there are. Since we know this at compile time, we may as well encode it in the type.

```ts

type Row = [BoardPosition, BoardPosition, BoardPosition, BoardPosition, BoardPosition, BoardPosition, BoardPosition, BoardPosition];

type BoardPositions = [Row, Row, Row, Row, Row, Row, Row, Row];

```

(This can be written a bit cleaner/fancier, but requires some Typescript magic)

packages/TorneloScoresheet/src/components/ChessBoard/ChessBoard.tsx
```diff
@@ -0,0 +1,36 @@
...
+  flipBoard: boolean;
+};
+
+const ChessBoard: React.FC<ChessBoardProps> = ({
+  boardPositions,
+  flipBoard,
+}) => {
+  if (flipBoard) {
+    boardPositions = boardPositions.reverse();
```
gestureleft, line 18: This reverses the array in-place. It is bad practice to mutate a React prop (as this can have unexpected behaviour for the parent).

Two options here - reverse the rendering with `flex-direction: 'row-reverse'`, or, take a copy:

```ts
{(flipBoard ? [...boardPositions].reverse() : boardPositions).map(...)};
```
(Or something similar)

packages/TorneloScoresheet/src/components/ChessBoard/style.ts
```diff
@@ -0,0 +1,18 @@
+import { StyleSheet } from 'react-native';
+import { colours } from '../../style/colour';
+
+export const styles = StyleSheet.create({
+  board: {
+    display: 'flex',
+    flexDirection: 'column',
+    height: 80 * 8,
+    width: 80 * 8,
```
gestureleft, line 9: Let's make this a constant that's exported from `ChessSquare.tsx` and imported here. That way, if we want to scale this, we can change it in one file.

`ChessSquare/style.tsx`:
```ts
export const CHESS_SQUARE_SIZE = 80;
```

gestureleft (changes requested):
Great stuff! Left some minor feedback

packages/TorneloScoresheet/src/components/ChessBoard/ChessBoard.tsx
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  return (
+    <View>
+      <RoundedView style={styles.board}>
+        {boardPositions.map((row, rowIdx) => (
+          <View style={styles.boardRow} key={'row-' + rowIdx.toString()}>
+            {row.map(square => (
+              <ChessSquare square={square} />
```
gestureleft, line 27: Chess square also needs a key

gestureleft (approved):

gestureleft (approved):
lgtm

PR #58, opened 2022-05-06
Updated pipeline file to allow for manual triggering

nocommentcode:
We can now manually trigger a github actions workflow,
see https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow for instructions


nocommentcode:
Will test this later, I think there's a better way

PR #59, opened 2022-05-07
Update styling to accomodate size of iPad 2nd Generation

gestureleft:
Also, fixed bug in iOS app entry-point and updated iOS dependencies

### DIFF
--- a/packages/TorneloScoresheet/src/pages/EnterPgn/style.ts
+++ b/packages/TorneloScoresheet/src/pages/EnterPgn/style.ts
@@ -1,17 +1,28 @@
-import { StyleSheet } from 'react-native';
+import { Dimensions, StyleSheet } from 'react-native';
+import { TOOLBAR_HEIGHT } from '../../components/Toolbar/style';
 import { colours } from '../../style/colour';
 import { primary } from '../../style/font';
+
+const { height } = Dimensions.get('screen');
 
 export const styles = StyleSheet.create({
   container: {
     display: 'flex',
-    alignItems: 'center',
+    justifyContent: 'space-between',
+    height: height - TOOLBAR_HEIGHT,
     backgroundColor: colours.white,
     paddingLeft: 58,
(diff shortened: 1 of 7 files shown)


nocommentcode (approved):
It's fine for now, but i really think we should strive to make the whole app responsive, even for the mvp, feels pretty obtuse only supporting 1 specfic version of 1 model of 1 brand of tablet.  
We can ask David and Simon what they thing  in our next meeting, maybe we can agree to only support tablets or somthing like that but loading the app on an ipad mini 2 instead of ipad mini 5 should not make the whole ui unusable...

gestureleft:
I agree - I'll update the PR to be responsive. May as well implement it properly while I'm looking at it

packages/TorneloScoresheet/src/pages/EnterPgn/style.ts
```diff
@@ -1,17 +1,28 @@
...
 import { primary } from '../../style/font';
 
+const { height } = Dimensions.get('screen');
+
 export const styles = StyleSheet.create({
   container: {
     display: 'flex',
-    alignItems: 'center',
+    justifyContent: 'space-between',
+    height: height - TOOLBAR_HEIGHT,
```
nocommentcode, line 12: Should we have this in a component like ResponsiveContainer because we'll be setting this height in each View right?

nocommentcode (approved):
Legend, i'll update GraphicalDisplay mode to also be responsive, should we perhaps have a ResponsiveContainer component that all our views can use?

packages/TorneloScoresheet/src/pages/EnterPgn/style.ts
gestureleft: Yeah we can definitely look into adding that abstraction. I think this is a fine first iteration though

nocommentcode (approved):
test

PR #61, opened 2022-05-09
Player move functionality

nocommentcode:
Updated the chess engine and the useGraphicalRecordingState() hook to be able to record moves.
Updated chess engine's unit test to thouroughly test the fenToBoardPositions() and makeMove() functions

### DIFF
--- a/packages/TorneloScoresheet/src/chessEngine/chessTsChessEngine.ts
+++ b/packages/TorneloScoresheet/src/chessEngine/chessTsChessEngine.ts
@@ -11,7 +11,7 @@ import {
   PlayerColour,
   PLAYER_COLOUR_NAME,
 } from '../types/ChessGameInfo';
-import { Piece, PieceType } from '../types/ChessMove';
+import { Piece, PieceType, PlySquares } from '../types/ChessMove';
 import { Result, succ, fail, isError } from '../types/Result';
 import { ChessEngineInterface } from './chessEngineInterface';
 
@@ -81,11 +81,42 @@ const startGame = (): [ChessBoardPositions, string] => {
 };
 
 /**
+ * Returns the board postion state given a fen
+ * @param fen the current state of the board
+ * @returns the board postions of the chess board
+ */
+const fenToBoardPositions = (fen: string): ChessBoardPositions => {
+  const game = new Chess(fen);
+  return gameToPeiceArray(game);
+};
+
+/**
+ * Processes a move given the starting fen and to and from positions
+ * @param startingFen the fen of the game state before the move
+ * @param plie the to and from positions of the move
+ * @returns the next fen if move is possible else null
+ */
+const makeMove = (
+  startingFen: string,
+  plySquares: PlySquares,
+): string | null => {
+  const game = new Chess(startingFen);
+  const result = game.force_move({ from: plySquares.from, to: plySquares.to });
+  if (result === null) {
+    return null;
+  }
+
+  return game.fen();
+};
+
+/**
  * The exported chess engine object which implements all the public methods
  */
 export const chessTsChessEngine: ChessEngineInterface = {
   parseGameInfo,
   startGame,
+  makeMove,
+  fenToBoardPositions,
 };
 
 // ------- Privates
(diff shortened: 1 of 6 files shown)


packages/TorneloScoresheet/src/chessEngine/chessTsChessEngine.ts
```diff
@@ -99,30 +127,33 @@ const gameToPeiceArray = (game: Chess): ChessBoardPositions => {
...
-      else {
-        row.push({
-          piece: mapEnginePeice(peice),
-          position: boardIndexToPosition(rowIdx, colIdx),
-        });
-      }
-    });
+  game
+    .board()
+    .reverse()
```
gestureleft, line 132: Let's not mutate a param.

`[...game.board()].reverse()`

packages/TorneloScoresheet/src/chessEngine/chessTsChessEngine.ts
```diff
@@ -99,30 +127,33 @@ const gameToPeiceArray = (game: Chess): ChessBoardPositions => {
...
+    .map((gameRow, colIdx) => {
```
gestureleft, line 133: Since we're not making a mapping transform here, I think it would be more semantic to call `forEach`. Unless you want to transform this into a mapping transform, which might be better? i.e. rather than:

```ts
const board = [];

game.forEach((el) => board.push(fn(el)))
```

You would do:
```ts
const board = game.map(el => fn(el))
```

packages/TorneloScoresheet/src/hooks/appMode/graphicalRecordingState.ts
```diff
@@ -1,19 +1,109 @@
...
+
+  // if move was not possible, return null
+  if (nextFen === null) {
+    return null;
+  }
+
+  // create next plie
+  const nextPlie: MovePlie = { startingFen: nextFen };
+
+  return nextPlie;
```
gestureleft, line 45:
```ts

return { startingFen: nextFen };

```

packages/TorneloScoresheet/src/hooks/appMode/graphicalRecordingState.ts
```diff
@@ -1,19 +1,109 @@
...
+};
+
+/**
+ * Processes a player's move give from and to positons
+ * Will Return the next board state and the move history array
+ * Will return null if move is imporssible
```
gestureleft, line 51: typo: impossible

packages/TorneloScoresheet/src/hooks/appMode/graphicalRecordingState.ts
```diff
@@ -30,13 +120,25 @@ export const makeUseGraphicalRecordingState =
...
     const goToTextInputFunc = (): void => {};
     const goToArbiterModeFunc = (): void => {};
+    const moveFunc = (pliePositions: PliePositions): void => {
+      const result = processPlayerMove(pliePositions, appModeState.moveHistory);
+      if (result !== null) {
+        const [board, moveHistory] = result;
+        setAppModeState({
+          ...appModeState,
+          board: board,
+          moveHistory: moveHistory,
```
gestureleft, line 130: Let's use the shorthand here: `{ a: a }` -> `{ a }`

gestureleft (changes requested):
It seems reasonable. Tests aren't passing though. Please see my feedback and get the tests passing

packages/TorneloScoresheet/src/hooks/appMode/graphicalRecordingState.ts
```diff
@@ -1,19 +1,109 @@
...
+    if (nextPlie === null) {
+      return null;
+    }
+
+    // return new state
+    const nextMove: ChessMove = {
+      moveNo: currentMove.moveNo + 1,
+      whiteMove: nextPlie,
+    };
+    moveHistory.splice(currentMoveIdx);
```
gestureleft, line 99: Also - don't think this if/else is necessary. Can't we use:
```ts
const moveIndex = currentMove.blackMove === undefined ? 'whiteMove' : 'blackMove';
currentMove[moveIndex].from = pliePositions.from;
...
```

I think the code would be much clearer without the massive if/else

gestureleft: You may need to use `'whiteMove' as const` and `'blackMove' as const` or something similar

packages/TorneloScoresheet/src/chessEngine/chessTsChessEngine.ts
```diff
@@ -99,30 +127,33 @@ const gameToPeiceArray = (game: Chess): ChessBoardPositions => {
...
+        }
+
+        // existing piece
+        else {
+          row.push({
+            piece: mapEnginePeice(peice),
+            position: boardIndexToPosition(rowIdx, colIdx),
+          });
+        }
+      });
```
gestureleft, line 153: Typo in piece.

Also, rather than an if else, I think you can do :

```ts
row.push({
  piece: piece === null ? null : mapEnginePiece(piece),
  position: ...
```

Usually prefer an expression over an if / else.

packages/TorneloScoresheet/src/hooks/appMode/graphicalRecordingState.ts
nocommentcode: yeh left it like that so it's clear what the difference between a white move update and a black move update is, but ive chagned it now i think its even clearer

gestureleft (changes requested):
Great changes! I noticed a few more small things

gestureleft (approved):
lgtm 💯

gestureleft (approved):
Even better!

gestureleft (approved):
Pog

PR #62, opened 2022-05-09
Implemented Drag and Drop

gestureleft:
This first iteration supports:
- animated dragging of elements across the screen
- having multiple drop targets and multiple draggables
- passing data between what is being dragged and the drop target it is being dropped on
- animating back to original position if drop target is hit


packages/TorneloScoresheet/src/types/ChessBoardPositions.ts
```diff
@@ -31,6 +31,74 @@ export type Row = '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8';
 export type Column = 'a' | 'b' | 'c' | 'd' | 'e' | 'f' | 'g' | 'h';
 export type Position = `${Column}${Row}`;
 
+// All the squares on a board
+export const BOARD_SQUARES: Position[] = [
```
nocommentcode, line 35: what's this for?

nocommentcode (approved):
looks good :)

packages/TorneloScoresheet/src/types/ChessBoardPositions.ts
gestureleft: Ah yep, this doesn't belong in this PR. Thanks!

nocommentcode (approved):
lgtm

PR #63, opened 2022-05-10
Table pairing mode

gwenllian1:
Implemented Table Pairing mode according to the figma. If the user clicks on a name, it asks them to confirm and takes them to the recording mode. Implemented by Kat and Llio.
capture
capture1


packages/TorneloScoresheet/android/gradle.properties
```diff
@@ -26,3 +26,4 @@ android.enableJetifier=true
 
 # Version of flipper SDK to use with React Native
 FLIPPER_VERSION=0.99.0
+org.gradle.jvmargs=-Xmx1536M --add-exports=java.base/sun.nio.ch=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED
```
nocommentcode, line 29: do you know what this is for?

packages/TorneloScoresheet/src/context/AppModeStateContext.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useContext, useState } from 'react';
 import { makeUseEnterPgnState } from '../hooks/appMode/enterPgnState';
 import { makeUsePairingSelectionState } from '../hooks/appMode/pairingSelectionState';
+import { makeUseEnterTablePairingState } from '../hooks/appMode/tablePairingState';
```
nocommentcode, line 4: should be called useTablePairingState, not useEnterTablePairingState (the enter comes from copy pasting makeUseEnterPgn i assume)

packages/TorneloScoresheet/src/pages/TablePairing/TablePairing.tsx
```diff
@@ -0,0 +1,139 @@
...
+      return;
+    }
+
+    const [board, fen] = chessEngine.startGame();
+    goToRecording(
+      tablePairingMode.pairing,
+      [{ moveNo: 1, whitePly: { startingFen: fen } }],
+      board,
+    );
+  };
```
nocommentcode, line 39: this type of logic is linked to transitioning between states and as such does not belong in the view, the code to start the game and build the movehistory should be in your useTablePairingState() hook and you can access the pairing from inside the hook too, and you should just call goToRecording() with no arguments from here.

packages/TorneloScoresheet/src/hooks/appMode/tablePairingState.ts
```diff
@@ -0,0 +1,49 @@
...
+      board: ChessBoardPositions,
+    ): void => {
+      setAppModeState({
+        mode: AppMode.GraphicalRecording,
+        pairing: pairing,
+        moveHistory: moveHistory,
+        board: board,
+      });
+    };
+
```
nocommentcode, line 47: There's no reason to take pairing, moveHistory and board as an argument to this function, it should have no argument and build the game from inside here, and it has access to the pairing via appModeState.pairing

packages/TorneloScoresheet/src/pages/TablePairing/TablePairing.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  const handleCancel = () => {
+    setShowSheet(false);
+  };
+
+  const displayPlayer = (
```
nocommentcode, line 45: harry has explained to me that all 'html' should be rendered in the final return on a component, so we shouldnt do stuff like this ( a method called displayPlayer that returns 'html'. If you find yourself wanting to call a function in the html, it should probably be its own separate component `PlayerCard`

packages/TorneloScoresheet/src/pages/TablePairing/TablePairing.tsx
```diff
@@ -0,0 +1,139 @@
...
+          <PrimaryText size={40} weight={FontWeight.Medium}>
+            TO DO TEAM
+          </PrimaryText>
+        </View>
+        <View style={styles.cardColumns}></View>
+      </View>
+    </TouchableOpacity>
+  );
+
+  const displaySheet = ({ pairing }: { pairing: ChessGameInfo }) => (
```
nocommentcode, line 87: same deal here i think a modal like this should be made as a general component as we will need to use a modal with 2 buttons multiple times throughout the project. Maybe for now we can just leave it like this

nocommentcode (changes requested):
Looking good, I just left a few comments, maybe wait till Harry has a look too beucase he generaly has good feedback !

packages/TorneloScoresheet/src/pages/TablePairing/TablePairing.tsx
ksov0001: We will be happy to implement this refactor next sprint, possibly a bit out of scope for this pull request.

packages/TorneloScoresheet/android/gradle.properties
ksov0001: Llio and I were both experiencing the following error: 

https://stackoverflow.com/questions/67782975/how-to-fix-the-module-java-base-does-not-opens-java-io-to-unnamed-module 

This fix worked for both of us - if we don't push this we need to add this line of code manually every time we pull. Hoping we can keep it.

nocommentcode (changes requested):
Cool looks good, 
On thing i forgot is that your tests are not passing because the TableParing mode test is nested inside another test, it should be on the outside. I would probably just remove that test because it isn’t actually testing anything, because it loads the `useTablePairingState()` hook, but the appModeContext will initially be set to EnterPgnMode so your hook will always return null, and thus you aren’t actually testing anything. 
To test it properly we need to be able to mock the appModeContext, which I just made a PR for. That pr will allow you to mock the context, which means the context will start off in whatever mode you want (in your case TablePairingMode ) instead of always starting off in EnterPgn mode, so you can then test what you need to test

nocommentcode (approved):
lgtm

gestureleft:
This PR is stale, opening a new one

PR #66, opened 2022-05-12
Implement Chess board drag and drop

gestureleft:


packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -32,14 +36,20 @@ const GraphicalRecording: React.FC = () => {
     goToArbiterMode();
   };
 
+  const position = graphicalRecordingMode?.board
```
nocommentcode, line 39: dont you think it makes more sense to have this logic inside the chessboard component itself and just have the chess board component take a `ChessBoardPosition` object instead?

nocommentcode (approved):
looks good, i just reckon we should shift the extracting of pieces to the chessboard component, since the requirement to represent the data like that comes from the chessboard component itself but up to you what do you think?

gestureleft:
> looks good, i just reckon we should shift the extracting of pieces to the chessboard component, since the requirement to represent the data like that comes from the chessboard component itself but up to you what do you think?

Yeah good point. I think it might be worth thinking about whether this way of representing the data will be useful in other contexts or just for the Board. Let's chat about it in person this arvo

PR #67, opened 2022-05-12
Update Table Pairing Mode UI 

gwenllian1:
capture2
capture1
Made some changes to the UI so they are now more similar to what is on the Figma.


nocommentcode (approved):
Looks much better!!

PR #71, opened 2022-05-17
Risk register update

gwenllian1:
updated the risk register to reflect the current state of the project
- Removed some irrelevant risks
- Added a couple of new risks (relating to FIDE approval and client communication)
- I also updated the risk matrix because it was taken from last year's project.


nocommentcode (approved):
Perfect!

nocommentcode (approved):
re run tests

PR #75, opened 2022-05-22
Action bar component

nocommentcode:
Created the `ActionButton` component and `ActionBar` component.

The `ActionButton` component takes the following parameters:
- text (to diplay on the button)
- icon (svg)
- onPress (a function triggered when the button is pressed)
- height (enum either single or double)

The `ActionBar` compnent takes in a list of actionButtonProps to display the buttons

### DIFF
--- a/packages/TorneloScoresheet/src/components/ActionButton/ActionButton.tsx
+++ b/packages/TorneloScoresheet/src/components/ActionButton/ActionButton.tsx
@@ -1,0 +1,50 @@
+import { View } from 'react-native';
+import { SvgProps } from 'react-native-svg';
+import { colours } from '../../style/colour';
+import PrimaryText, { FontWeight } from '../PrimaryText/PrimaryText';
+import RoundedView from '../RoundedView/RoundedView';
+import React from 'react';
+import { TouchableOpacity } from 'react-native-gesture-handler';
+import { styles } from './style';
+
+export enum ButtonHeight {
+  SINGLE,
+  DOUBLE,
+}
+export type ActionButtonProps = {
+  Icon: React.FC<SvgProps>;
+  onPress: () => void;
+  text: string;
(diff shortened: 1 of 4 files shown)


packages/TorneloScoresheet/src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <RoundedView
+        style={[
+          styles.buttonContainer,
+          {
+            height: (buttonHeight + 1) * SINGLE_UNIT_HEIGHT,
+            width: SINGLE_UNIT_HEIGHT,
+          },
+        ]}>
+        <View
+          style={{ height: '70%', display: 'flex', justifyContent: 'center' }}>
```
gwenllian1, line 40: just double checking if it's ok for the style to be in here instead of the style file?

packages/TorneloScoresheet/src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+          <Icon height={40} fill={colours.white}></Icon>
```
gestureleft, line 41: This should be a self-closing tag

gestureleft (changes requested):
Nice! I agree with Llio, and have suggested one more change

packages/TorneloScoresheet/src/components/ActionButton/ActionButton.tsx
nocommentcode: it is very sinful and an accident

gestureleft: We'll discuss your punishment at the next standup ;)

gestureleft (approved):
Nice

PR #76, opened 2022-05-22
Graphical recording buttons

nocommentcode:
Added the `ActionBar` component to graphical recording mode

- Added the button icons to the assets folder
- Added a style.ts file for `GraphicalRecordingMode`
- Added the `ActionBar` component to graphical recording mode
- Wrapped the `ChessBoard` and `ActionBar` components in flex box
- Linked the flip board button as it is easy to do

image


packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -21,14 +91,18 @@ const GraphicalRecording: React.FC = () => {
...
-        <View>
-          <ChessBoard
-            positions={graphicalRecordingMode.board}
-            onMove={onMove}
-            flipBoard={
-              graphicalRecordingMode.currentPlayer === PlayerColour.Black
-            }
-          />
+        <View style={styles.mainContainer}>
+          <View style={{ height: 100 }}>
```
gestureleft, line 95: Inline styling here as well - let's put this in `style.ts`

gestureleft (changes requested):
Nice! One small change

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
nocommentcode: this is just a placeholder though, will be replaces by the player cards once we implement it

gestureleft: True! My bad

gestureleft: Should probably have a TODO/FIXME there but doesn't matter - I'll approve now

gestureleft (approved):
🔥

PR #77, opened 2022-05-23
Update Chess Engine behaviour

gestureleft:
Specifically, updated the behaviour of the `move` function. Please see the commit messages for details, but this patch basically makes the behaviour less unexpected, and removes the requirement for having a "dummy" move in the history at the start of a game. :^)

### DIFF
--- a/packages/TorneloScoresheet/__tests__/chessEngine.test.ts
+++ b/packages/TorneloScoresheet/__tests__/chessEngine.test.ts
@@ -279,6 +279,13 @@ test('testStartGame', () => {
   expect(countPeices(board, PlayerColour.Black)).toStrictEqual(16);
 });
 
+test('initialFen', () => {
+  const startingFen = chessEngine.initialFen();
+  expect(startingFen).toStrictEqual(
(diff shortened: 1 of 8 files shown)


nocommentcode (changes requested):
I think this implementation is wrong, 
This change stores the resulting fen in each ply instead of the starting fen. This means that a ply is no longer a stand alone representation of the move, as you need to access the previous move in order to 'understand' a ply. 
The reason for the way it has been implemented is that we want the Ply to hold all the info required for that ply and nothing else, and be able to operate on that ply without having to find the previous one (which requires additional checks depending on if this is white's/black's move)
If you think about it semmantically, a ply represents the state of the board (starting fen), and a move that altered the board (squares). By changing it the way you did, whilst it might make more sense to you now, it will make the following more confusing:
- Editing a move -> When we will edit a move, we will need to get the fen before that move, to setup the board in the way it was before the move was made, so we want the Ply to store the starting fen and the squares of the move
- Skipping a move -> when we skip a move, we will need to have the 'resultingFen' in that skipped move which doesnt make sense, instead of just copying the starting fen over to the next ply
Let me know what you think tho

gestureleft:
Sure! I will go back to using starting fen instead of resultant fen. Makes sense!

gestureleft:
Superseded by #81

PR #79, opened 2022-05-26
Option Popup Component

nocommentcode:
Replaced the `ConfirmPlayerSheet` with a generic component that we can now use everywehre (I need it for another story)
The `OptionPopup` Component prompts the user to click on a button in a modal.
It takes the following arguments:
- message -> the message to prompt the user
- onCancel -> the function called when exiting the modal
- visible -> a state boolean variable
- options -> an array of type Option, which contains text, icon and onPress function

Each option in the modal can have a text, an icon, or both.
Demo:
 
image


packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
```diff
@@ -0,0 +1,66 @@
+import { TouchableOpacity, View } from 'react-native';
+import { SvgProps } from 'react-native-svg';
+import RoundedView from '../RoundedView/RoundedView';
+import Sheet from '../Sheet/Sheet';
+import { styles } from './style';
+import { colours } from '../../style/colour';
+import PrimaryText, { FontWeight } from '../PrimaryText/PrimaryText';
+import React from 'react';
+
+type Option = {
```
gwenllian1, line 10: Instead of option would something like popupContent be better? If you prefer option I'm happy to approve it :)

B-Gunawan: I'd say option makes sense, since we're not defining the JSX content of option buttons here.

packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
```diff
@@ -0,0 +1,66 @@
...
+                    height={55}
+                  />
+                )}
+                {option.text && (
+                  <PrimaryText style={styles.buttonText}>
+                    {option.text}
+                  </PrimaryText>
+                )}
+              </RoundedView>
+            </TouchableOpacity>
```
B-Gunawan, line 58: Can we convert this to a generic component? i.e. `IconTextButton`

```TypeScript
IconTextButton: React.FC <{
  icon: React.FC<SvgProps>,
  text: String,
  onPress: () => void,
  rounded: boolean,
}>
```

B-Gunawan: So you'd end up with
```TypeScript
{options.map((option, i) => (
  <IconTextButton
    icon={option.icon}
    text={option.text}
    option={option.action} // or option.onPress
  />
))}
```
Makes the component a bit more readable.

packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
```diff
@@ -0,0 +1,66 @@
...
+  onCancel,
+}) => {
+  return (
+    <Sheet dismiss={onCancel} visible={visible}>
+      <View style={styles.mainContainer}>
+        <PrimaryText
+          style={styles.messageText}
+          weight={FontWeight.Bold}
+          size={30}
+          colour={colours.darkenedElements}>
```
B-Gunawan, line 35:
``` typescript
   colour={colours.darkenedElements}
>
{{ message }}
```

whitespace for readability. not sure if this is a prettier issue though.

packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
```diff
@@ -0,0 +1,66 @@
...
+              <RoundedView style={styles.button}>
+                {option.icon && (
+                  <option.icon
+                    fill={colours.white}
+                    stroke={colours.white}
+                    color={colours.white}
+                    height={55}
+                  />
+                )}
+                {option.text && (
```
B-Gunawan, line 52: It'd also be nice to add some spacing between the button icon and button text.

B-Gunawan (changes requested):

packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
B-Gunawan: Alternatively, we can add an optional `icon` prop to `PrimaryButton`. 
```JSX
<PrimaryButton
  icon={...}
  style={...}
  labelStyle={...}
  onPress={...}
  label="...."
/>
```

B-Gunawan (changes requested):

packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
B-Gunawan: Alternatively, Add an optional `icon` prop to `PrimaryButton`
```JSX
<PrimaryButton
  style={...}
  label={...}
  labelStyle={...}
  icon={...}
  iconStyle={...}
  onPress={...}
/>
```

packages/TorneloScoresheet/src/components/TextIconButton/TextIconButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  TouchableOpacityProps,
+} from 'react-native';
+import { SvgProps } from 'react-native-svg';
+import { colours } from '../../style/colour';
+import PrimaryText from '../PrimaryText/PrimaryText';
+import RoundedView from '../RoundedView/RoundedView';
+import { styles } from './style';
+
+type TextIconButtonParms = {
+  Icon?: React.FC<SvgProps>;
```
gestureleft, line 15: This is an unusual API design. I think it might be more appropriate to accept the icon as a string and render an 'react-native-vector-icons/MaterialIcons' `Icon` accordingly? (See our `IconButton` component). Curious to hear your thoughts, because it probably doesn't matter massively either way, just seems unnecessarily generic/abstract or smt?

packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
```diff
@@ -0,0 +1,56 @@
...
+import PrimaryText, { FontWeight } from '../PrimaryText/PrimaryText';
+import React from 'react';
+import TextIconButton from '../TextIconButton/TextIconButton';
+
+type Option = {
+  icon?: React.FC<SvgProps>;
+  text?: string;
+  onPress: () => void;
+};
+type OptionPopupParams = {
```
gestureleft, line 15: Our naming convention would have this be `OptionPopupProps`. I'll update our CONTRIBUTING guide to say this

packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
```diff
@@ -0,0 +1,56 @@
...
+  message: string;
+  options: Option[];
+  visible: boolean;
+  onCancel: () => void;
+};
+
+const OptionPopup: React.FC<OptionPopupParams> = ({
```
gestureleft, line 22: To me it makes more sense to name this `OptionSheet`, since that is the terminology we have used for the `Sheet` component. (otherwise we can rename `Sheet` to `Popup` - either works)

gestureleft (changes requested):
Good change overall - I left some minor notes

packages/TorneloScoresheet/src/components/TextIconButton/TextIconButton.tsx
nocommentcode: This resticts us to Google Font Material icons though, I'm using the icons straight from the figma file, those can be downloaded in svg form. This is the same approach used for our `PieceAssets` as well as the `ActionButton` components, I think its fine this way as SVG images are superior to all others, and this way we have the freedom to use any icon

packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
nocommentcode: Makes sense i'd just never heard the term sheet (I normally use the term 'modal' i think that comes from boostrap) Lets stick with sheet throughout the project

packages/TorneloScoresheet/src/components/TextIconButton/TextIconButton.tsx
scd: FYI, the main Tornelo system currently has the same policy with downloading svg.

packages/TorneloScoresheet/src/components/OptionPopup/OptionPopup.tsx
scd: FYI, modal is used as terminology in the main Tornelo system. Of course you are free to choose whatever you want.

gestureleft (approved):
Nice

PR #80, opened 2022-05-26
Pawn promotion

nocommentcode:
Implemented the Pawn Promotion Functionality
When the user drags a Pawn into the final row on either side, they will be promted to select their desired promotion
Demo:

https://user-images.githubusercontent.com/26475724/170444448-6e0c5a8d-27f9-453e-83c5-e0255a54b066.mp4

### DIFF
--- a/packages/TorneloScoresheet/__tests__/chessEngine.test.ts
+++ b/packages/TorneloScoresheet/__tests__/chessEngine.test.ts
@@ -949,3 +949,81 @@ describe('makeMove', () => {
     });
   });
 });
+
+// ----- chessEngine.isPawnPromotion() ----
+
+describe('isPromotion', function () {
+  const testCases = [
+    {
+      name: 'legal move non-promotion',
+      fen: 'rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1',
+      san: 'e4',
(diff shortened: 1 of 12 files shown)


packages/TorneloScoresheet/__tests__/chessEngine.test.ts
```diff
@@ -949,3 +949,67 @@ describe('makeMove', () => {
...
+      san: 'cxb8',
+      move: { from: 'c7', to: 'b8' },
+      promotion: true,
+    },
+  ];
+
+  testCases.forEach(function (testCase) {
+    it(testCase.name, function () {
+      expect(
+        chessEngine.isPawnPromotion(testCase.fen, testCase.move as PlySquares),
```
gestureleft, line 1011: Should this be `as MoveSquares`?

gestureleft (changes requested):
Looks good! It isn't compiling (see the comment I've left), but other than that looks great!

packages/TorneloScoresheet/__tests__/chessEngine.test.ts
nocommentcode: Yeh weird as the test was still running and passing? Not sure how that's possible

gestureleft (approved):
LGTM

PR #81, opened 2022-05-26
Refractor move history

nocommentcode:
Change how we store our history,
Instead of storing it in a `ChessMove` array which has whitePly and blackPly, 
We store just the plys directly in an array
This is much much cleaner as we don't have to do the checks for whos turn it is
This will also make it so much easier to undo and skip moves


gestureleft (approved):
Great changes! This is a much clearer and more straightforward implementation

PR #84, opened 2022-06-16
Auto skip turn

nocommentcode:
When a player moves their piece twice in a row,
It is assumed that they intended to skip the other player's turn and register another move

This is acheived by the `isOtherPlayersPiece()` function in the chess.ts engine.
This function detects if the user is selecting a piece from the opposite color of the current turn.

When a player moves a piece the flow is as follows:
- Check if the move is a pawn promotion
- If promotion -> prompt user for promotion choice
- Determine if the piece moved is the wrong color (ie player is attempting to auto skip)
- If auto skip -> record `SkipPly` for current player and a `MovePly` with the desired move
- If not auto skip -> record a `MovePly` with the desired move for the current player

Demo: (Notice how when I undo, i have to press undo twice for moves that were skip since we are also undoing the `SkipPly`.

https://user-images.githubusercontent.com/26475724/173999250-2242fa4f-473b-4f97-9316-244040d88d19.mp4

### DIFF
--- a/packages/TorneloScoresheet/__tests__/chessEngine.test.ts
+++ b/packages/TorneloScoresheet/__tests__/chessEngine.test.ts
@@ -1050,3 +1050,45 @@ describe('Skip Turn', () => {
     });
   });
 });
+
+// ---- chessEngine.isOtherPlayersPiece() ----
+describe("is other player's piece", () => {
+  const testCases = [
+    {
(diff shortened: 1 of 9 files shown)


gestureleft (approved):
Great work! 🔥 💯 👍 😎 🎉

PR #87, opened 2022-07-03
Fixed zIndex not working for pieces being dragged

gestureleft:
Previously the piece being dragged would sometimes render under other pieces.

https://user-images.githubusercontent.com/57308619/177023511-3d1b8218-ea3c-439b-ab8b-3c8ca573418e.mov


nocommentcode (approved):
This is so simple I can’t believe we didn’t think of it before 🙄🙄 gg

PR #89, opened 2022-07-03
Arbiter Mode button

gwenllian1:
Created button to enter arbiter mode. It is displayed in the toolbar when the app is in player mode. It currently has no functionality so just throws an error when it's clicked on but we are implementing the pin input later this week.
image
image


gwenllian1:
Closing this because Kat and I are making some changes.

PR #91, opened 2022-07-05
Add Arbiter Button in Toolbar and Arbiter Pin Popup

ksov0001:
Base functionality of entering Arbiter mode from player mode. 

Button is visible in Player mode (and not Arbiter mode). 

https://user-images.githubusercontent.com/41656704/177318881-23e4ad4f-7296-4eaa-818e-e5f770b5d682.mp4

TODO 
We will do another pull request to make the pin sheet input more user friendly. 
Will also make Pin into a component.


packages/TorneloScoresheet/node_modules/react-native-pincode-input-component/src/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'react-native-pincode-input-component';
```
gestureleft, line 1: It's a bit strange to check `node_modules` into the repository - since this is basically a build artifact. Do we need this here for some reason?

gwenllian1: yeah, we had to make it, it wasn't there by default so we figured it was easier to just push it instead of everyone needing to do it individually

gwenllian1:
We have decided to use a different package

PR #92, opened 2022-07-05
Updated dependencies

gestureleft:
Updated to the all of the latest dependencies :))

We had several vulnerabilities, so this fixes those

We moved from "@testing-library/react-hooks" to "@testing-library/react
", since the react one now supports the API that we first went to
react-hooks for, and it supports react 18, which react-hooks does not


gwenllian1 (approved):

gwenllian1 (approved):

gwenllian1 (approved):

gestureleft:
This is breaking the CI for some reason. I guess we will hold off on updating the dependencies for a while

PR #93, opened 2022-07-08
Select the winner and enter player signature

gwenllian1:
https://user-images.githubusercontent.com/64511606/177945020-ecff47bf-8b15-483f-a2d3-dbac1f8d2640.mp4

- when the end game is picked the player should select who won
- the player is prompted for a signature 
- when the signature is entered the app switches to result mode (currently unimplemented)

Future TODO: 
- make it so that both players must sign
- optional arbiter signature


packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -224,18 +229,24 @@ const GraphicalRecording: React.FC = () => {
...
                   text: fullName(graphicalRecordingMode.pairing.players[1]),
-                  onPress: handleSelectWinner,
+                  onPress: () =>
+                    handleSelectWinner(
+                      graphicalRecordingMode.pairing.players[1],
+                    ),
                 },
               ]}
               visible={showEndGame}
               onCancel={cancelSelection}
```
nocommentcode, line 253: Would be nice to put these options at the top as a variable called winnerSelectionOptions to be consistent (the action buttons and promotion buttons are both under the buttom parameters comment on line 85), otherwise our view gets super messy with all these sheet options!

packages/TorneloScoresheet/src/components/Signature/Signature.tsx
```diff
@@ -0,0 +1,72 @@
...
+            showTitleLabel={false}
+            viewMode={'portrait'}
+          />
+          <View style={styles.buttonArea}>
+            <PrimaryButton
+              style={styles.buttonStyle}
+              onPress={() => {
+                confirmSign();
+                onConfirm();
+              }}
```
nocommentcode, line 57: Can we have confirm on the right and reset on the left? Generally that is the standard way right?

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -65,8 +71,16 @@ const GraphicalRecording: React.FC = () => {
...
+    setShowSignature(true);
+    setShowEndGame(false);
+    setSelectedWinner(player);
+  };
+
   const cancelSelection = () => {
+    setShowSignature(false);
     setShowEndGame(false);
+    setSelectedWinner(undefined);
   };
```
nocommentcode, line 84: Also I think these functions should go under the button params section, maybe write a comment on line 159 called 'Button functions' or something and move your button functions to join the others

nocommentcode (changes requested):
Looks really good love the signature, just a few things:
- Where are we saving the image atm? I think we should have `saveSignature` and `loadSignature` functions in util.storage.ts where it saves and loads the signature image.
- We should keep our function names consistent, we are mixing it up between 'handle___' and 'on___' all over our pages, maybe we should fix this in a separate PR
- I would at least change the `handleConfirm` function name in this PR to something more clear like `handleConfirmWinner` or something so we know which function does what because the GraphicalRecording Page is getting really big and messy
- We should pass through the winner in the `goToEndGame()` call as we will need it to render the results screen
- Bug: When I click on end game and select a player, when I’m in the signature popup, the board and all other components disappear for some reason?

packages/TorneloScoresheet/src/components/Signature/Signature.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const sign = createRef<any>();
+
+const confirmSign = () => {
+  sign.current.saveImage();
+};
+
+const resetSign = () => {
+  sign.current.resetImage();
+};
```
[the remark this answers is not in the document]
gestureleft: Hmm yeah, this is strange. We should be using `useRef` and all of this should be in the component I believe

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
gwenllian1: How badly do you want these not to be here because I can't just have it as a variable (it complains that graphicalRecordingMode could be undefined) and from some investigation, I think it would be a pain to implement it elsewhere because of the way option sheet is set up

nocommentcode: hmmm i guess it doesnt matter that much its just to keep it clean, you can do something like this though:
image

gwenllian1: oh cool I didn't know I could do that (my lack of react experience is showing), I fixed everything else so I'll update the pr. (Also stop doing code on your birthday, do and do something fun :) )

gwenllian1:
Where are we saving the image atm? I think we should have saveSignature and loadSignature functions in util.storage.ts where it saves and loads the signature image.

I'm not currently doing anything with the image, to use the saved image we just need to define an onSaveEventFunction and it is returned as a base64 encoded string (example here: https://github.com/RepairShopr/react-native-signature-capture/blob/master/Example/src/ExampleApp.js) but I wasn't sure what we actually wanted to do with it so I thought that might be another ticket for when we know more about where it should be sent/stored etc.

We should pass through the winner in the goToEndGame() call as we will need it to render the results screen

I was going to do this for the result screen ticket because the pr was getting kind of big

nocommentcode (approved):
sweet lgtm

PR #95, opened 2022-07-11
Function naming

gwenllian1:
Updated function naming to match the following convention:

  onEvent={handleEvent}


nocommentcode (approved):
Thanks for this :)

PR #96, opened 2022-07-12
Update transition from Graphical Recording Mode to Result Display Mode

nocommentcode:
**This PR has been changed**
_Instead of storing using AsyncStorage, we will instead pass the base64 images of signatures to the ResultDisplay app mode state_

When the signature is confiremed, it is saved as a base64 image, and then the `GraphicalRecording` page will call the hook's `goToEndGame` method and pass through the winner and the signature of the current player.
As @gwenllian1 mentioned, in the future we might also require the other player and the arbiter to also sign, we can confirm how this should work with David when we see him next and update it accordingly

In the hook's `goToEndGame` function, the code currently calls a `generatePgnFromHistory` function which currently returns an empty string, we will need to implement this function in a separate PR.

This PR also extended the functionality of the end game feature by allowing the user to select a draw

### DIFF
--- a/packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
+++ b/packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
@@ -179,11 +179,14 @@ const GraphicalRecording: React.FC = () => {
   };
 
   //Button Functions
-  const handleConfirmWinner = () => {
+  const handleConfirmWinner = (signature: string) => {
     if (!graphicalRecordingMode || !goToEndGame) {
       return;
     } else {
-      goToEndGame();
+      goToEndGame({
+        winner: selectedWinner?.color ?? null,
+        signature: signature,
+      });
     }
   };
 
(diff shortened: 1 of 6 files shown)


gwenllian1:
This is really cool! Will it let us store multiple signatures because we are going to need to get the signatures of both players and optionally the arbiter.

gwenllian1:
Just out of curiosity what are the benefits of passing the base64 image to results display mode compared to storing in AsyncStorage? 

With regards to the draw option is the idea that the offer draw button is just to record the fact that a draw was offered but doesn't end the game? or will both the offer draw and end game buttons allow users to end the game on a draw?

nocommentcode:
> Just out of curiosity what are the benefits of passing the base64 image to results display mode compared to storing in AsyncStorage?

Actually there is no need to store it in storage, the storage is only needed if you want the data to still exist after closing the app and reopening it, so we need it to store the pgn url, but for the signature, its only needed in the next screen so we can just pass it into the state like we do for all other data (eg we dont store the current `ChessGameInfo` in storage, we just store it in the state (in RAM)

> With regards to the draw option is the idea that the offer draw button is just to record the fact that a draw was offered but doesn't end the game? or will both the offer draw and end game buttons allow users to end the game on a draw?

Yes so I think the draw offer will just record the fact a draw offer was made at that point in time, but it will not end the game, when the player declares the game as ended, they can record the result as a draw then.

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -156,8 +156,16 @@ const GraphicalRecording: React.FC = () => {
             width: '100%',
           },
         },
+        {
+          text: 'draw',
```
gwenllian1, line 160: Can this have a capital d?

PR #98, opened 2022-07-15
Arbiter pin

ksov0001:
Added button to Enter Arbiter Mode in the toolbar with a sheet (note: only visible in Player Mode). 
Added Pin entry option to the Enter Arbiter Mode sheet in the toolbar.
Added Verify Pin button (currently not performing any validation) which moves the app into Arbiter Game Mode.

https://user-images.githubusercontent.com/41656704/179174139-48e0279f-ce8e-45f6-99c8-b44ec893f906.mp4


packages/TorneloScoresheet/src/components/Toolbar/ToolbarButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const ToolbarButton: React.FC<ToolbarButtonProps> = ({
+  Icon,
+  colour,
+  display,
+  onPress,
+}) => {
+  return (
+    <TouchableOpacity onPress={() => onPress()} style={{}}>
+      <RoundedView
```
nocommentcode, line 22: We actually already have a `ActionButton` component that we use for the action buttons, 
It accepts a svg file for the icon, 
text for the name,
a function to trigger on press

I think we should resuse that component instead of creating a new one since its basically the same thing,
The only edit needed is to change the styling, (right now the button has a blue background and the icon and text are white) you can add a boolen parameter called 'invert_colours' or something that would render a white button with blue interiors

packages/TorneloScoresheet/src/style/images.ts
```diff
@@ -32,3 +33,4 @@ export const ICON_HASTAG = ICON_HASHTAG_ASSET;
 export const ICON_REPEAT = ICON_REPEAT_ASSET;
 export const ICON_UNDO = ICON_UNDO_ASSET;
 export const ICON_CLOCK = ICON_CLOCK_ASSET;
+export const ICON_ARBITER_MODE = ICON_ARBITER_MODE_ASSET;
```
nocommentcode, line 36: If we resuse the component i'm talking about, we should import the gears icon instead of the whole button image here, ie import ICON_GEARS instead of ICONS_ARBITER_MODE

packages/TorneloScoresheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -30,17 +35,30 @@ const colourForMode: Record<AppMode, ColourType> = {
...
 };
 
+const arbiterModeDisplay: Record<AppMode, 'none' | undefined> = {
+  [AppMode.EnterPgn]: 'none',
+  [AppMode.PariringSelection]: 'none',
+  [AppMode.TablePairing]: undefined,
+  [AppMode.GraphicalRecording]: undefined,
+  [AppMode.ResultDisplay]: undefined,
+};
+
```
nocommentcode, line 53: Any reason for mapping to 'none' | undefined? why not map to a boolean true/false for each mode?

packages/TorneloScoresheet/src/components/Pin/Pin.tsx
```diff
@@ -0,0 +1,46 @@
...
+        ref={ref}
+        {...props}
+        // Use `caretHidden={false}` when users can't paste a text value, because context menu doesn't appear
+        value={value}
+        onChangeText={setValue}
+        cellCount={CELL_COUNT}
+        rootStyle={styles.codeFieldRoot}
+        keyboardType="number-pad"
+        textContentType="oneTimeCode"
+        renderCell={({ index, symbol, isFocused }) => (
```
nocommentcode: should be `renderCell={({ index, symbol, isFocused } : {index: number, symbol: string, isFocused: boolean}) => (` to remove the error

packages/TorneloScoresheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -72,7 +74,9 @@ const Toolbar: React.FC = () => {
       <Sheet
         title="Enter Pin"
         dismiss={() => setShowArbiterSheet(false)}
-        visible={showArbiterSheet}></Sheet>
+        visible={showArbiterSheet}>
+        <Pin></Pin>
```
nocommentcode: use `<Pin />` format for consistancy

packages/TorneloScoresheet/src/components/Pin/Pin.tsx
```diff
@@ -19,27 +25,32 @@ const Pin = () => {
...
+              key={index}
+              style={[styles.cell, isFocused && styles.focusCell]}
+              onLayout={getCellOnLayoutHandler(index)}>
+              {symbol || (isFocused ? <Cursor /> : null)}
+            </Text>
+          )}
+        />
+      </SafeAreaView>
+      <View style={styles.verifyButtonArea}>
+        <PrimaryButton onPress={() => onPress()} label="Verify" />
```
nocommentcode, line 54: We should pass the value of the pin into this function no?
Also you can just write `onPress={onPress}` , no need to declare another function

packages/TorneloScoresheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -62,6 +68,28 @@ const Toolbar: React.FC = () => {
...
+
+  const graphicalRecordingState = useGraphicalRecordingState();
+  const goToArbiterGameFromGraphicalRecording =
+    graphicalRecordingState?.[1]?.goToArbiterGameMode;
+
+  const tablePairingState = useTablePairingState();
+  const goToArbiterGameFromTablePairing =
+    tablePairingState?.[1]?.goToArbiterGameMode;
+
+  const handleVerify = () => {
```
nocommentcode, line 80: even if we are not handling the checking logic yet, maybe this function should take the pin value as an input and maybe console log it or something witha  todo comment saying we need to implement the check

packages/TorneloScoresheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -62,6 +68,28 @@ const Toolbar: React.FC = () => {
...
+        return;
+      }
+      goToArbiterGameFromGraphicalRecording();
+    } else {
+      if (!goToArbiterGameFromTablePairing) {
+        return;
+      }
+      goToArbiterGameFromTablePairing();
+    }
+  };
```
nocommentcode, line 93: This will get messy as we add more and more AppModes here, maybe we should store a dictionary mapping the game mode to the transition function?

ie at the top:
  `const appModeToAribiterMode = Record<AppMode, () => void>`
and then in handleVerify() :
 `appModeToArbiterMode[appModeState]()`

also if we change the appMode to ArbiterGameMode, we will loose all the data in the PlayerMode states, so either we need to pass all the data to the ArbiterGameMode or we need to find a way to store it, I think we need to discuss this all together to decide what the best way of handling the arbiter mode will be, maybe we should have a separate arbiter mode per app mode ie have a TablePairingArbiterMode, GraphicalRecordingArbiterMode, ... etc

packages/TorneloScoresheet/src/hooks/appMode/arbiterGameState.ts
```diff
@@ -0,0 +1,27 @@
...
+  ): (() => ArbiterGameStateHookType | null) =>
+  (): ArbiterGameStateHookType | null => {
+    const [appModeState, setAppModeState] = useContext(context);
+
+    if (appModeState.mode !== AppMode.ArbiterGame) {
+      return null;
+    }
+
+    //    const goToArbiterGameFunc =
+    //    makegoToArbiterGame(setAppModeState);
```
nocommentcode, line 24: Whats this

packages/TorneloScoresheet/src/hooks/appMode/graphicalRecordingState.ts
```diff
@@ -155,7 +155,11 @@ export const makeUseGraphicalRecordingState =
       });
     };
     const goToTextInputFunc = (): void => {};
-    const goToArbiterModeFunc = (): void => {};
+    const goToArbiterModeFunc = (): void => {
+      setAppModeState({
+        mode: AppMode.ArbiterGame,
+      });
+    };
```
nocommentcode, line 162: We need to pass through the graphicalRecordingState's data here, otherwise we will loose it! (Pairing, currentPlayer, moveHistory etc...

packages/TorneloScoresheet/src/hooks/appMode/tablePairingState.ts
```diff
@@ -11,6 +11,8 @@ type TablePairingStateHookType = [
   TablePairingMode,
   {
     goToRecording: (currentPlayer: PlayerColour) => void;
+
+    goToArbiterGameMode: () => void;
```
nocommentcode, line 15: Same here we need to pass through the state's data or it will be lost forever :(

nocommentcode (changes requested):
Looks cool, 
however we need to decide how we will handle the ArbiterMode, because the way you have implemented it now i see two problems:

1) All the data from the previous AppMode is lost, so if were in graphical recording for example, the pairing, current player move history etc is lost, so essentialy the game would reset once we exit the arbiter mode

2) Right now you have the same aribiter mode for all app modes, i think we should have a dedicated one for each mode, so we should have a TablePairingArbiterMode, a GraphicalRecordingArbiterMode etc becuase each of these pages will have different displays and buttons depending on where we are in the flow, and if we have these different modes, we can solve problem 1 by passing the current state's data into the arbiter mode (this will be necessary anyways since the arbiter mode will need to access this data and maybe change it (think checking the legality of a move while in GraphicalRecordingArbiterMode for example)

packages/TorneloScoresheet/src/components/Toolbar/ToolbarButton.tsx
nocommentcode: Edit: You'll also need a visible property to show/hide it, it can be optional and shown by default to not affect the current invocations

packages/TorneloScoresheet/src/components/Pin/Pin.tsx
gwenllian1: If we do onPress={onPress} we get an error. We were holding off on passing the value of the pin anywhere until we talk to David about how they want it stored because we assume it will need to be encrypted or something which is outside the scope of this ticket.

packages/TorneloScoresheet/src/components/Toolbar/Toolbar.tsx
gwenllian1: I couldn't get the dictionary to work but I turned it into a switch statement so it's a bit neater. We would probably have a maximum of 4 possible changes to arbiter mode (tablePairing, graphicalrecording, resultdisplay, and textrecording) so hopefully it shouldn't offend Harry too much

nocommentcode:
@gwenllian1  I finaly got the dictionary to work took me forever, I also tried the onPress param thingy and its working for me!
Everything else looks great!

nocommentcode (approved):

nocommentcode:
@gwenllian1 I also changed the styling of the toolbar to make it thinner.
Let me know what you think i know it isnt the same as figma but i feel like the way it was was way too fat!

Old:
old

New:
new

PR #99, opened 2022-07-17
Generate pgn When ending the game

nocommentcode:
The PR implements the logic to generate the pgn at the end of the game.
The pgn will be generated before prompting the user for a signature so that if there are any errors, the user will be notified before signing.

### DIFF
--- a/packages/TorneloScoresheet/src/chessEngine/chessEngineInterface.ts
+++ b/packages/TorneloScoresheet/src/chessEngine/chessEngineInterface.ts
@@ -1,6 +1,6 @@
 import { BoardPosition } from '../types/ChessBoardPositions';
-import { ChessGameInfo } from '../types/ChessGameInfo';
-import { PieceType, MoveSquares } from '../types/ChessMove';
+import { ChessGameInfo, PlayerColour } from '../types/ChessGameInfo';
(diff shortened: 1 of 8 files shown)


packages/TorneloScoresheet/src/chessEngine/chessEngineInterface.ts
```diff
@@ -67,6 +67,19 @@ export type ChessEngineInterface = {
    * @returns if the move is targeting the opposite player's piece
    */
   isOtherPlayersPiece: (fen: string, move: MoveSquares) => boolean;
+
+  /**
+   * Generates the PGN of the game
+   * @param originPgn the pgn of the even with the headers
```
gestureleft, line 73: What does "even" mean in this sentence?

gestureleft (changes requested):
Nice! Looks really good! I've left one small comment

packages/TorneloScoresheet/src/chessEngine/chessEngineInterface.ts
nocommentcode: event ** good spot

gestureleft (approved):
LGTM

PR #100, opened 2022-07-18
Edited Styling of toolbar to fix styling when in arbiter mode

nocommentcode:
I broke the toolbar styling while in arbiter mode when i changed it in the previous PR,

**Before:**
old

**Now:**
new

I also removed the unused ArbiterGameState as it was unused and throwing an error
We will need to create a state for each type of arbiter mode


gwenllian1:
In regards to the AppModeStateContext, that was my mistake for not updating it because I made arbiter modes for the graphical recording and table pairing modes (ArbiterGraphicalRecording and ArbiterTablePairing) so the app mode state context should probably include those.

nocommentcode:
> In regards to the AppModeStateContext, that was my mistake for not updating it because I made arbiter modes for the graphical recording and table pairing modes (ArbiterGraphicalRecording and ArbiterTablePairing) so the app mode state context should probably include those.

Yes I think we will implement these 2 custom hooks when we crease the pages for them, there will be:
- useArbiterTablePairingState
- useArbiterGraphicalRecordingState
- useArbiterResultDisplayState

We'll create those as we make the ArbiterTablePairing, ArbiterGraphicalRecording and ArbiterResultDisplay pages in the coming sprints

gwenllian1 (changes requested):
In the figma the Tornelo logo is on the left not in the middle, would it be possible to make it match that?

gwenllian1:
How did you break the styling? The first one looked the it matched the figma styling: 
image

nocommentcode:
> How did you break the styling? The first one looked the it matched the figma styling: image

I changed the padding to make the game mode one narrower, so the arbiter one became super narrow.
Yeh the figma has the arbiter one with the tornelo branding on the left, however that was before the new design for the player mode toolbar had the branding in the middle.

I changed it to be consitant i feel like it makes more sense to have these 2 toolbars:
new
blue

Instead of these 2:
old
blue

On top of consitancy, i just realised that in arbiter mode for graphical recording, table pairing and results display, we are probably going to want a 'Player mode' button in the same place as the 'arbiter mode' button as a quick way for the arbiter to exit arbiter mode and go back to the previous player mode screen?

What do you think

gwenllian1 (approved):

gwenllian1:
Yeah, that makes sense, consistency across the app would be good and it makes the branding more prominent having it in the centre. It's a lot better the way you want to do it

PR #101, opened 2022-07-19
Toggle draw

gwenllian1:
https://user-images.githubusercontent.com/64511606/179705184-9d17d146-b1a1-42ab-9dac-135b9397f648.mp4

Implemented the offer draw functionality. It either adds a draw offer or removes it depending on if there already is one for the move. 

The draw offer function in the graphical recording hook takes the index of the move and updates the draw offer status. Currently, I'm just passing the last move in the list but in the future when we implement the 'go back and edit any move' functionality we should be able to pass the index of any selected move. 

For the video, I removed the placeholder in graphical recording mode because my emulator doesn't show the full move card with it there.

### DIFF
--- a/packages/TorneloScoresheet/__tests__/useAppModeState.test.ts
+++ b/packages/TorneloScoresheet/__tests__/useAppModeState.test.ts
@@ -131,6 +131,7 @@ describe('graphical recording moving', () => {
             move,
             type: PlyTypes.MovePly,
             player: PlayerColour.White,
+            drawOffer: false,
           },
         ],
       });
@@ -149,6 +150,7 @@ describe('graphical recording moving', () => {
         startingFen: 'rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1',
         move: { from: 'a1', to: 'a5' } as MoveSquares,
         type: PlyTypes.MovePly,
+        drawOffer: false,
       },
     ];
 
@@ -171,6 +173,7 @@ describe('graphical recording moving', () => {
             startingFen:
               'rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1',
             move: { from: 'a1', to: 'a5' } as MoveSquares,
+            drawOffer: false,
           },
           {
             moveNo: 1,
@@ -178,6 +181,7 @@ describe('graphical recording moving', () => {
             type: PlyTypes.MovePly,
             startingFen: originFen,
             move,
+            drawOffer: false,
           },
         ],
       });
@@ -209,6 +213,7 @@ describe('undoing last move', () => {
         startingFen: 'rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1',
         type: PlyTypes.MovePly,
         move: { from: 'a1', to: 'a5' } as MoveSquares,
+        drawOffer: false,
       },
     ];
     const graphicalState = generateGraphicalRecordingState(moveHistory);
@@ -233,6 +238,7 @@ describe('undoing last move', () => {
         type: PlyTypes.MovePly,
         startingFen: 'rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1',
         move: { from: 'a1', to: 'a5' } as MoveSquares,
+        drawOffer: false,
       },
       {
         moveNo: 1,
@@ -240,6 +246,7 @@ describe('undoing last move', () => {
         startingFen: 'rnbqkbnr/pppppppp/8/R7/8/8/PPPPPPPP/1NBQKBNR b Kkq - 1 1',
         move: { from: 'h8', to: 'h5' } as MoveSquares,
         type: PlyTypes.MovePly,
+        drawOffer: false,
       },
     ];
     const graphicalState = generateGraphicalRecordingState(moveHistory);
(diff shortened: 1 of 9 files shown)


nocommentcode (changes requested):
Nice work, will be super easy to implement draw offers for any move since you used the index :)

Just a few things:
- Do we need the confirmation popup? skip turn and undo don't have a popup, they just happen straight away, since this is a toggle its super easy to undo so i reckon you can just directly add the draw offer once the button is pressed
- The way it is styled right now is a bit strange, when you add the draw offer, it moves the text down and the padding is not even on the top and bottom, i reccomend having a placeholder view with the same height as the icon that is shown when the icon is not shown and is hidden when the icon is shown, that way when you toggle the draw the move text will not move at all, also we need to add padding on the bottom to match the top. This will probably mean the move card will be too big now, but we can probably make the move number at the top skinnier to compensate and keep the total card the same height

gwenllian1:
I made it so that the formatting doesn't change when a draw offer is added, but I had to make the move number part a lot smaller.

https://user-images.githubusercontent.com/64511606/179905236-505b1a6e-583a-4448-963a-5284f82e4d7e.mp4

nocommentcode (changes requested):
The styling is perfect now! I reckon the smaller move counter looks better than before tbh.

Just 1 other thing i realised, when you click draw offer, you add a draw offer to the previous move. 
I think it should be for the move which is currently being recorded. 

Right now the toggle draw button is toggling draw on the last move that way recorded, when it should be the move yet to be recorded, this is going to be a little tricky to implement as it requires toggling a draw on a move that does not yet exist... I think either we change the `ChessPly` to have optional arguments everywhere except for the draw offer so you can record a draw before a move is recorded, but this is going to be messy as we're going to need null checks everywhere...
Another option would be to store a state 'nextMoveIsDraw' as a boolean, and once that move is recorded we pass through its draw status ( i think this is probably the way to go)

Let me know what you think

gestureleft:
I don't think we should be toggling draw on moves that haven't been made yet? In Chess you offer a draw after making a move, so it doesn't make sense to toggle draw on a move that hasn't been made yet

nocommentcode:
> I don't think we should be toggling draw on moves that haven't been made yet? In Chess you offer a draw after making a move, so it doesn't make sense to toggle draw on a move that hasn't been made yet

Oh are we sure you only offer a draw after you play your own move?

gestureleft:
> Under [FIDE](https://en.wikipedia.org/wiki/FIDE) [rules](https://en.wikipedia.org/wiki/Rules_of_chess), a draw should be offered after making the move and before pressing the [game clock](https://en.wikipedia.org/wiki/Game_clock)

nocommentcode:
> From Wikipedia:
> 
> >

Oh sweet in that case ignore my comment @gwenllian1 , im happy with the PR unless @gestureleft wants to review it as well

gwenllian1:
@nocommentcode if @gestureleft doesn't want me to make any changes can you please approve it?

gestureleft:
No issues from my end. (Haven't had a chance to read it, but I'm sure it's fine :)) )

nocommentcode (approved):
LGTM

nocommentcode:
The tests failed because you need to add the draw offer to the expected results.
Also it would be good if you can add unit tests for the draw offer stuff.
All hook methods should be tested, there's plenty of examples in the useAppModeStateTests on how to write tests for the hook methods !

gestureleft (approved):
LGTM

gestureleft:
Tests are still failing. (Failing on my machine also) Are they passing on yours @gwenllian1

gwenllian1:
image
It's failing on the test cases I wrote because it's receiving an anonymous function. I think it might be because my return value is calling a function and I haven't mocked that but I'm not sure how to do that 
image
@gestureleft @nocommentcode

gestureleft:
@gwenllian1 No need to mock it. Please see an example of how to test this sort of thing here: https://github.com/gestureleft/TorneloScoresheet/commit/b51e163fc1a8797c9377d9f5aacab97e26f948dc

Let me know if you have questions :))

nocommentcode:
> It's failing on the test cases I wrote because it's receiving an anonymous function. I think it might be because my return value is calling a function and I haven't mocked that but I'm not sure how to do that

Yes @gwenllian1  Its because you're calling a function in set appmode state
image

Theres no need for this function since we are already checking that we are in graphical recording mode above
image

Instead you can simply just do this
image

And then 2 of the failling the tests pass :)

The other 2 erorrs are due to a bug, you forgot to changer the hook method called in ''offer draw black move', and 'remove draw offer black move' , you also dont need to change the board state, ie should be:
image

and 

image

gestureleft:
@nocommentcode  We definitely want to use a function here in the state setter, since in this case the new state is a function of the previous state. If we use your approach we may introduce bugs since React batches state updates. See here for an explanation: https://beta.reactjs.org/apis/usestate#updating-state-based-on-the-previous-state

nocommentcode:
> @nocommentcode We definitely want to use a function here in the state setter, since in this case the new state is a function of the previous state. If we use your approach we may introduce bugs since React batches state updates. See here for an explanation: https://beta.reactjs.org/apis/usestate#updating-state-based-on-the-previous-state

wdym we've been calling set state with an object all over the state hooks, we just need to make sure to make copies of the objects right?
So it should be:
image

gestureleft:
> wdym we've been calling set state with an object all over the state hooks, we just need to make sure to make copies of the objects right? So it should be: image

In most cases it's fine to update directly. In cases like this where the new state is defined in terms of the previous state, it is more semantic to define it as a function, and means that it's impossible for us to trip up on React's batching. So there's not really any reason to not define it as a function, whereas there is reasons why we would want to define it as a function.

gestureleft (approved):
Nice!

gestureleft (approved):
LGTM :)))

PR #102, opened 2022-07-20
Minor fixes

gestureleft:
This PR is a bit of a mixed bag:

- Fixed vulnerability
- Made the linter happy
- Fixed bug where the layout was wrong
- Updated the styling of the arbiter button
- Updated the API of `ActionButton`

The arbiter button styling is to make it more consistent with the "help button", and to make the toolbar smaller - I'm open to iterations here, but the previous design was too big imo. Plus, the "arb" text didn't really make the button clearer - what does "arb" mean? I might be wrong, but I don't think it will be clear at all to users that it is short for "arbiter mode". So now, we put that information in the sheet that immediately greets users when they press it:

new arbiter button

New arbiter sheet


nocommentcode (approved):
Looks good although you'll need to comit again to fix the merge conflicts (the `generatePgnFromHistory` from your branch can go, it has been replaces in #99 

As for the arbiter button i agree, looks better like this, but i think in that case we'll need a new icon, because the gears icon will be confused for settings. I think maybe a flight attendant style icon or maybe a padlock would be suitable to enter arbiter mode?

nocommentcode (approved):
Sweet

PR #105, opened 2022-07-25
Implemented Lock app on exit

gestureleft:
If the app is exited or moved to the background, we lock the app down

This PR also introduces the global modal stack context, to keep track of which modal to show. This is required because React Native only supports showing one modal at a time. We get around this by managing our own stack of modals with this new context.

### DIFF
--- a/packages/TorneloScoresheet/App.tsx
+++ b/packages/TorneloScoresheet/App.tsx
@@ -8,6 +8,8 @@ import { AppModeStateContextProvider } from './src/context/AppModeStateContext';
 import { ErrorContextProvider } from './src/context/ErrorContext';
 import Main from './src/pages/Main';
 import { GestureHandlerRootView } from 'react-native-gesture-handler';
+import LockAppOnExit from './src/components/LockAppOnExit/LockAppOnExit';
+import { ModalStackContextProvider } from './src/context/ModalStackContext';
 
 const App = () => {
   const isDarkMode = useColorScheme() === 'dark';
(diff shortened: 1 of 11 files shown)


nocommentcode (approved):
GG,
I think the `Pin` component should handle the pin checking, so we can resue it.
Instead of onPress the pin should take a onCorrectPinEntry arg,
Not related to this pr tho

gestureleft:
> GG, I think the `Pin` component should handle the pin checking, so we can resue it. Instead of onPress the pin should take a onCorrectPinEntry arg, Not related to this pr tho

I've noticed that this doesn't work if there is already a different modal open.

Also, it shouldn't show for arbiter mode views. So I'll fix those and re-request a review. Soz, this is what happens when u rush I guess haha 😬

nocommentcode:
Poor form.
I tried defeating it a few ways, it works against everything i tried
One thing i was thiking of is gestures,
In ios, when you do a gesture you can start to see other open apps right?
but i dont think it counts as an actual exit from the app.
Can you try see if you do a partila gesture could you see what was open in a chome tab for exampel?

gestureleft:
> Poor form. I tried defeating it a few ways, it works against everything i tried One thing i was thiking of is gestures, In ios, when you do a gesture you can start to see other open apps right? but i dont think it counts as an actual exit from the app. Can you try see if you do a partila gesture could you see what was open in a chome tab for exampel?

Yep - partially peaking into another app locks it :))

I've updated the PR to only look out when not in arbiter mode

Also updated the PR to introduce modal stack context

nocommentcode (approved):
lgtm

PR #106, opened 2022-07-25
Fix PGN Result Header Bug.

nocommentcode:
The result header of the pgn will now also include the result.


gwenllian1 (changes requested):
There's a typo in pgnSuccess

gwenllian1 (approved):

PR #107, opened 2022-07-25
fixed sheet close and pin input bugs, kat and llio

gwenllian1:
Fixed import that meant the x button was not working on android devices.
Fixed the pin button error, however the keyboard only appears for ios users (Harry has verified it appears)

https://user-images.githubusercontent.com/64511606/180726332-0030fdeb-62d9-4b55-904b-41c960040919.mp4


gestureleft (approved):
LGTM

PR #109, opened 2022-07-26
Move options

gestureleft:
Pressing and holding on a MoveCard ply now brings up a Move Options sheet.

See below for a demo:

https://user-images.githubusercontent.com/57308619/180930880-8ef6caed-c8dd-4196-bf4a-d5a98b10db6c.mov

### DIFF
--- a/packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
+++ b/packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import Sheet from '../../components/Sheet/Sheet';
+import { colours } from '../../style/colour';
+import { ICON_CLOCK, ICON_HALF } from '../../style/images';
+import { PlayerColour } from '../../types/ChessGameInfo';
+import MoveOption from './MoveOption';
+import Icon from 'react-native-vector-icons/FontAwesome';
+import MaterialIcon from 'react-native-vector-icons/MaterialIcons';
+import { useGraphicalRecordingState } from '../../context/AppModeStateContext';
+
+export type EditingMove = {
+  colour: PlayerColour;
+  moveIndex: number;
+};
+
+type MoveOptionsSheetProps = {
+  editingMove: EditingMove | undefined;
+  dismiss: () => void;
+};
+
+const MoveOptionsSheet = ({ editingMove, dismiss }: MoveOptionsSheetProps) => {
+  const recordingState = useGraphicalRecordingState();
+  const actions = recordingState?.[1];
+
+  const toggleDraw = actions?.toggleDraw;
+
+  const handleDrawOffer = () => {
+    console.log(editingMove);
+    if (!toggleDraw || !editingMove) return;
+    toggleDraw(
+      editingMove.moveIndex * 2 +
+        (editingMove.colour === PlayerColour.Black ? 1 : 0),
+    );
+    dismiss();
+  };
+
+  return (
+    <Sheet
+      visible={Boolean(editingMove)}
+      dismiss={dismiss}
+      title="Move options">
+      <MoveOption
+        onPress={() => undefined}
+        colour={colours.primary}
+        icon={<ICON_CLOCK color={colours.white} />}
+        label="Record time"
+      />
+      <MoveOption
+        onPress={handleDrawOffer}
+        colour={colours.primary}
+        icon={<ICON_HALF color={colours.white} />}
+        label="Draw offer"
+      />
+      <MoveOption
+        onPress={() => undefined}
+        colour={colours.primary}
+        icon={<Icon name="edit" size={35} color={colours.white} />}
+        label="Edit move"
+      />
+      <MoveOption
+        onPress={dismiss}
+        colour={colours.secondary}
+        icon={<MaterialIcon name="cancel" size={35} color={colours.white} />}
+        label="Cancel"
+      />
+    </Sheet>
+  );
+};
+
+export default MoveOptionsSheet;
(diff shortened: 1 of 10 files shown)


packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
```diff
@@ -0,0 +1,70 @@
...
+    );
+    dismiss();
+  };
+
+  return (
+    <Sheet
+      visible={Boolean(editingMove)}
+      dismiss={dismiss}
+      title="Move options">
+      <MoveOption
```
nocommentcode, line 42: any reason we're not passing the function here? will it be part of the other ticket?

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
```diff
@@ -0,0 +1,70 @@
...
+};
+
+type MoveOptionsSheetProps = {
+  editingMove: EditingMove | undefined;
+  dismiss: () => void;
+};
+
+const MoveOptionsSheet = ({ editingMove, dismiss }: MoveOptionsSheetProps) => {
+  const recordingState = useGraphicalRecordingState();
+  const actions = recordingState?.[1];
```
nocommentcode, line 23: I feel like this function should be passed through by the graphical recording state, so all the functions are declared in one spot, ie we shouldnt be accesing the state hook here

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
```diff
@@ -0,0 +1,70 @@
...
+        onPress={() => undefined}
+        colour={colours.primary}
+        icon={<Icon name="edit" size={35} color={colours.white} />}
+        label="Edit move"
+      />
+      <MoveOption
+        onPress={dismiss}
+        colour={colours.secondary}
+        icon={<MaterialIcon name="cancel" size={35} color={colours.white} />}
+        label="Cancel"
```
nocommentcode, line 64: The cancel button isnt working for me? I dont understand why though

nocommentcode (changes requested):
Looks good,
I'm not sure about the sheet styling though, i feel like the title should be front and center in a bigger font, plus with some of the sheets the titles are long so it makes the sheet extra wide.
I feel like we should have 2 fields: a title and a message field, where the title is a short small font in the top left and the title is a larger centered text field at the top of the sheet:

I think this looks worse than before


This one is ok


This one did not get the change so we need to at least be consistant

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
nocommentcode: Especially since the draw function is already defined in the graphical recording page, so if there is extra logic before calling the hook we need to change it in 2 places

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
nocommentcode: Actually none of the buttons work for me. Even the console logging doesnt show up so i think the click is not being registered. Im on android

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
gestureleft: The record time function? I think it's not implemented yet right? I thought that was your ticket

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
gestureleft: I'm a bit confused by this, they are all defined in the one spot : bottom of `AppModeStateContext.tsx`. This component will be using a few of the functionalities provided by `useGraphicalRecordingState`: The record time function, the draw offer function and the edit move function. The point of putting this stuff in the hook is so that it's decoupled from the component logic and can basically act as a "service" to the UI logic. I'm not saying your wrong, I'm just confused by what you're saying, so I'm happy to have a chat about this over zoom or something to clarify if you want?

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
gestureleft: Okay I think I've fixed this now. Let me know if it's still not working

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
nocommentcode: What i mean is that maybe there is extra logic that needs to be performed before calling the hook, which will need to be implemented both in GraphicalRecording page ( for the action buttons) and this component, so it might be better to pass through a function into this component that is used both by the action buttons in GraphicalRecording and this popup. But actually right now the toggle draw code is just 1 line of code so we can just go with your implementation for now

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
nocommentcode: yep no worries i will implement it

packages/TorneloScoresheet/src/pages/GraphicalRecording/MoveOptionsSheet.tsx
nocommentcode: yeh all good working now :)

nocommentcode (approved):
LGTM

PR #110, opened 2022-07-27
Signature page

gwenllian1:
Implemented signatures from both players. I chose to do them one at a time for the following reasons:
- it prevents one player from accidentally signing the wrong area
- it stops a player from accidentally editing the signature of the other player
- the UI looks less cluttered
- I tried a version with both signatures on the same popup but I didn't have room to fit two different confirm and reset buttons so if one player wanted to re-do their signature they both would have had to.

Also happy to take any suggestions to improve the design

https://user-images.githubusercontent.com/64511606/181193329-06c75552-1ca1-4697-b312-b25b81f76237.mp4

### DIFF
--- a/packages/TorneloScoresheet/--verify/_logs/2022-07-26T02_00_44_640Z-debug-0.log
+++ b/packages/TorneloScoresheet/--verify/_logs/2022-07-26T02_00_44_640Z-debug-0.log
@@ -1,0 +1,36 @@
+0 verbose cli [
+0 verbose cli   'C:\\Program Files\\nodejs\\node.exe',
+0 verbose cli   'C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js',
+0 verbose cli   '--cache',
+0 verbose cli   '--verify'
+0 verbose cli ]
+1 info using npm@8.5.0
+2 info using node@v16.14.2
+3 timing npm:load:whichnode Completed in 0ms
+4 timing config:load:defaults Completed in 2ms
+5 timing config:load:file:C:\Program Files\nodejs\node_modules\npm\npmrc Completed in 2ms
+6 timing config:load:builtin Completed in 3ms
+7 timing config:load:cli Completed in 2ms
+8 timing config:load:env Completed in 0ms
+9 timing config:load:file:C:\Users\lliov\Documents\FIT4002\scoresheet-app\packages\TorneloScoresheet\.npmrc Completed in 1ms
+10 timing config:load:project Completed in 6ms
+11 timing config:load:file:C:\Users\lliov\.npmrc Completed in 0ms
+12 timing config:load:user Completed in 0ms
+13 timing config:load:file:C:\Users\lliov\AppData\Roaming\npm\etc\npmrc Completed in 1ms
+14 timing config:load:global Completed in 1ms
+15 timing config:load:validate Completed in 1ms
+16 timing config:load:credentials Completed in 0ms
+17 timing config:load:setEnvs Completed in 1ms
+18 timing config:load Completed in 16ms
+19 timing npm:load:configload Completed in 16ms
+20 timing npm:load:setTitle Completed in 1ms
+21 timing config:load:flatten Completed in 2ms
+22 timing npm:load:display Completed in 3ms
+23 verbose logfile C:\Users\lliov\Documents\FIT4002\scoresheet-app\packages\TorneloScoresheet\--verify\_logs\2022-07-26T02_00_44_640Z-debug-0.log
+24 timing npm:load:logFile Completed in 7ms
+25 timing npm:load:timers Completed in 0ms
+26 timing npm:load:configScope Completed in 0ms
+27 timing npm:load Completed in 27ms
+28 verbose exit 0
+29 timing npm Completed in 217ms
+30 verbose code 1
--- a/packages/TorneloScoresheet/__tests__/useAppModeState.test.ts
+++ b/packages/TorneloScoresheet/__tests__/useAppModeState.test.ts
@@ -967,7 +967,7 @@ describe('goToResultDisplayFromGraphicalRecording', () => {
     const graphicalStateHook = renderCustomHook(useGraphicalRecordingState);
     const result = {
       winner: PlayerColour.White,
-      signature: ['base 64 image', 'base 64 image'],
+      signature: { 0: 'base 64 image', 1: 'base 64 image' },
       gamePgn: 'game pgn',
     };
 
@@ -1104,7 +1104,7 @@ describe('goToResultDisplayFromGraphicalRecording', () => {
     const graphicalStateHook = renderCustomHook(useGraphicalRecordingState);
     const result = {
       winner: PlayerColour.Black,
-      signature: ['base 64 image', 'base 64 image'],
+      signature: { 0: 'base 64 image', 1: 'base 64 image' },
       gamePgn: 'game pgn',
     };
 
@@ -1241,7 +1241,7 @@ describe('goToResultDisplayFromGraphicalRecording', () => {
     const graphicalStateHook = renderCustomHook(useGraphicalRecordingState);
     const result = {
       winner: null,
-      signature: ['base 64 image', 'base 64 image'],
+      signature: { 0: 'base 64 image', 1: 'base 64 image' },
       gamePgn: 'game pgn',
     };
 
--- a/packages/TorneloScoresheet/__tests__/useAppModeState.test.ts
+++ b/packages/TorneloScoresheet/__tests__/useAppModeState.test.ts
@@ -967,7 +967,7 @@ describe('goToResultDisplayFromGraphicalRecording', () => {
     const graphicalStateHook = renderCustomHook(useGraphicalRecordingState);
     const result = {
       winner: PlayerColour.White,
-      signature: 'base 64 image',
+      signature: ['base 64 image', 'base 64 image'],
       gamePgn: 'game pgn',
     };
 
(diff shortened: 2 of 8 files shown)


packages/TorneloScoresheet/--verify/_logs/2022-07-26T02_00_44_640Z-debug-0.log
```diff
@@ -0,0 +1,36 @@
...
+8 timing config:load:env Completed in 0ms
+9 timing config:load:file:C:\Users\lliov\Documents\FIT4002\scoresheet-app\packages\TorneloScoresheet\.npmrc Completed in 1ms
+10 timing config:load:project Completed in 6ms
+11 timing config:load:file:C:\Users\lliov\.npmrc Completed in 0ms
+12 timing config:load:user Completed in 0ms
+13 timing config:load:file:C:\Users\lliov\AppData\Roaming\npm\etc\npmrc Completed in 1ms
+14 timing config:load:global Completed in 1ms
+15 timing config:load:validate Completed in 1ms
+16 timing config:load:credentials Completed in 0ms
+17 timing config:load:setEnvs Completed in 1ms
```
nocommentcode, line 23: I think we should add --verify/logs to git ignor

packages/TorneloScoresheet/__tests__/useAppModeState.test.ts
```diff
@@ -967,7 +967,7 @@ describe('goToResultDisplayFromGraphicalRecording', () => {
     const graphicalStateHook = renderCustomHook(useGraphicalRecordingState);
     const result = {
       winner: PlayerColour.White,
-      signature: 'base 64 image',
+      signature: ['base 64 image', 'base 64 image'],
```
nocommentcode, line 970: Might be better to have a Record<PlayerColor, string>

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -52,11 +52,13 @@ const GraphicalRecording: React.FC = () => {
...
   const [, showError] = useError();
   const [showEndGame, setShowEndGame] = useState(false);
-  const [showSignature, setShowSignature] = useState(false);
+  const [showFirstSignature, setShowFirstSignature] = useState(false);
+  const [showSecondSignature, setShowSecondSignature] = useState(false);
   const goToEndGame = graphicalRecordingState?.[1].goToEndGame;
   const [selectedWinner, setSelectedWinner] = useState<
     undefined | Player | null
   >(undefined);
+  const [playerSignatures, setPlayerSignatures] = useState<string[]>([]);
```
nocommentcode, line 61: I think it's better to have a Record<PlayerColor, string|undefined> here too, it will be undefined untill a signiture is stored.

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -171,18 +173,32 @@ const GraphicalRecording: React.FC = () => {
...
 
+  const handleConfirmFirstSignature = (signatureInput: string) => {
+    setPlayerSignatures(playerSignatures => [
+      ...playerSignatures,
+      signatureInput,
+    ]);
+    setShowFirstSignature(false);
+    setShowSecondSignature(true);
+  };
+
```
nocommentcode, line 201: This works but I think we can come up with something slightly more elegant, if we have a record<PlayerColor, string|undefined> for the signatures, I think we can avoid having 2 separate handle confirm functions and 2 separate Signature components

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -197,13 +213,14 @@ const GraphicalRecording: React.FC = () => {
     setPgn(pgnResult.data);
 
     // if pgn can be generated -> prompt user for signature
-    setShowSignature(true);
+    setShowFirstSignature(true);
```
nocommentcode, line 216: So here we would call handleConfirmWinner()

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -171,18 +173,32 @@ const GraphicalRecording: React.FC = () => {
...
+      if (signatureArray.length == 2) {
+        setPlayerSignatures(() => []);
+        goToEndGame({
+          winner: selectedWinner?.color ?? null,
+          signature: signatureArray,
+          gamePgn: pgn,
+        });
+      }
     }
   };
```
nocommentcode, line 191: And then in here we would be checking which of the signature Record still has an undefined for the signature, set the Signature Component's player to the one who needs to sign, then show the signature popup, the signature component's onConfirm function will be this function again, where the signature is saved and the same check will be performed again for the other player, and once both signature records are not undefined, we can proceed to call the hooks function

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -287,12 +304,22 @@ const GraphicalRecording: React.FC = () => {
...
+            visible={showSecondSignature}
             onCancel={handleCancelSelection}
             winnerName={(selectedWinner && fullName(selectedWinner)) ?? null}
             onConfirm={handleConfirmWinner}
-            whitePlayer={graphicalRecordingMode.pairing.players[0]}
-            blackPlayer={graphicalRecordingMode.pairing.players[1]}
+            white={graphicalRecordingMode.pairing.players[0]}
+            black={graphicalRecordingMode.pairing.players[1]}
+            currentPlayer={graphicalRecordingMode.pairing.players[1]}
           />
```
nocommentcode, line 332: Same deal here, I think we can clean this up by having only 1 signature component, where we change the current player to change who is signing

packages/TorneloScoresheet/src/components/Signature/Signature.tsx
```diff
@@ -57,10 +59,10 @@ const Signature: React.FC<SignatureProps> = ({
             weight={FontWeight.Bold}
             size={40}
             colour={colours.darkenedElements}>
-            {`${fullName(whitePlayer)} \n `}
-            {`${resultText(winnerName, fullName(whitePlayer))}`}
+            {`${fullName(white)} \n `}
+            {`${resultText(winnerName, fullName(white))}`}
```
nocommentcode, line 63: I think it could be good to have whoever is singing have a bold name, and the other player's name is not bold, that way it's extra clear who is supposed to be signing

nocommentcode (changes requested):
Looks good, I agree, I prefer the flow where each player signs 1 by 1.
I just left some feedback with the way you implemented the 2 signatures, It works fine the way it is but I think we can make it cleaner by only using 1 signature component and doing a bit of logic to swap the current player around, but honestly its not a big deal since its not like in the future we are going to be adding more and more signatures...

packages/TorneloScoresheet/src/components/Signature/Signature.tsx
gwenllian1: The text is already bold, and I think it's pretty clear who is meant to be signing, if you think it's necessary I can make the 'signature of player' part bigger? It also might be confusing to bold the player in the result part because it's meant to display who won not who is signing.
image

nocommentcode: hmmm yeh good point actually that might be conbfusing

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -52,13 +52,20 @@ const GraphicalRecording: React.FC = () => {
   const [pgn, setPgn] = useState('');
   const [, showError] = useError();
   const [showEndGame, setShowEndGame] = useState(false);
-  const [showFirstSignature, setShowFirstSignature] = useState(false);
-  const [showSecondSignature, setShowSecondSignature] = useState(false);
+  const [showSignature, setShowFirstSignature] = useState(false);
```
nocommentcode, line 55: setShowSignature **

packages/TorneloScoresheet/src/types/ChessGameInfo.ts
```diff
@@ -15,7 +15,7 @@ export type ChessGameInfo = {
 export type ChessGameResult = {
   winner: PlayerColour | null;
   gamePgn?: string;
-  signature: string[];
+  signature: { 0: string | undefined; 1: string | undefined };
```
nocommentcode, line 18: Why not Record<PlayerColor, string> ?
i think at this point the signature can never be undefined right?
Because there is no option to skip the signing step,

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -175,30 +182,37 @@ const GraphicalRecording: React.FC = () => {
...
+      graphicalRecordingMode.currentPlayer == PlayerColour.White
+        ? !signatureRecord[0]
+          ? (signatureRecord[0] = signatureInput)
+          : (signatureRecord[1] = signatureInput)
+        : !signatureRecord[1]
+        ? (signatureRecord[1] = signatureInput)
+        : (signatureRecord[0] = signatureInput);
+
+      setPlayerSignatures(signatureRecord);
+      console.log(signatureRecord);
```
nocommentcode, line 202: can get rid of this now

nocommentcode (changes requested):
Just 2 things, looks much better though :)

packages/TorneloScoresheet/src/types/ChessGameInfo.ts
gwenllian1: If I do that I get this error 
image

nocommentcode: Yep because in the graphical page the signature can still be undefined, so we need to build a new one where we have a default if the signature is undefined (it should never be at that point though because were already checking on line 203
image

nocommentcode (approved):
Lgtm

PR #111, opened 2022-08-01
Adding time to moves

nocommentcode:
We can now add time to the move
The time is relative to the start of the game

Changes made: 
- ChessPly now has an optional gameTime property
- MoveCard now displays clock icon if a move has a time attached
- GraphicalRecordingState now stores the time of the game starting
- new TimePickerSheet component allows editing a time for a move
- Added set time functionality to GraphicalRecording state
- Updated GraphicalRecodring page and the move options sheet to allow recording time 

https://user-images.githubusercontent.com/26475724/182061531-71b125d4-9260-4330-89ab-e624e2bd9089.mp4

### DIFF
--- a/packages/TorneloScoresheet/src/components/TimePickerSheet/TimePickerSheet.tsx
+++ b/packages/TorneloScoresheet/src/components/TimePickerSheet/TimePickerSheet.tsx
@@ -28,7 +28,7 @@ const TimePickerSheet: React.FC<TimePickerSheetProps> = ({
 
   // refresh game time whenever it changes
   useEffect(() => {
-    setDate(new Date(1, 1, 1, gameTime.hours + 1, gameTime.minutes, 0));
+    setDate(new Date(1, 1, 1, gameTime.hours, gameTime.minutes, 0));
   }, [gameTime]);
 
   return (
--- a/packages/TorneloScoresheet/src/components/TimePickerSheet/TimePickerSheet.tsx
+++ b/packages/TorneloScoresheet/src/components/TimePickerSheet/TimePickerSheet.tsx
@@ -1,0 +1,67 @@
+import React, { useEffect, useState } from 'react';
+import { View } from 'react-native';
+import DatePicker from 'react-native-date-picker';
+import { GameTime } from '../../types/ChessMove';
+import PrimaryButton from '../PrimaryButton/PrimaryButton';
+import Sheet from '../Sheet/Sheet';
+import { styles } from './style';
+
+export type TimePickerSheetProps = {
+  gameTime: GameTime;
+  visible: boolean;
+  dismiss: () => void;
+  setGameTime: (gameTime: GameTime | undefined) => void;
+};
+
+const TimePickerSheet: React.FC<TimePickerSheetProps> = ({
+  gameTime,
+  visible,
+  dismiss,
+  setGameTime,
+}) => {
+  // set day/month/year not important
+  // date should be + 1 on hours for it to display properly,
+  // this is taken into account when setting the date
+  const [date, setDate] = useState(
+    new Date(1, 1, 1, gameTime.hours, gameTime.minutes, 0),
+  );
+
+  // refresh game time whenever it changes
+  useEffect(() => {
+    setDate(new Date(1, 1, 1, gameTime.hours + 1, gameTime.minutes, 0));
+  }, [gameTime]);
+
+  return (
+    <Sheet dismiss={dismiss} visible={visible} title={'Set move time'}>
+      <View style={styles.container}>
+        <DatePicker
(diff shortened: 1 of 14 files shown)


gestureleft (changes requested):
I like it! I left a suggestion about how we store timing related stuff, let me know what you think!

packages/TorneloScoresheet/src/types/AppModeState.ts
```diff
@@ -41,6 +41,7 @@ export type TablePairingMode = {
 
 export type GraphicalRecordingMode = {
   mode: AppMode.GraphicalRecording;
+  startTime: GameTime;
```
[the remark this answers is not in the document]
nocommentcode: Yeh good point a game lasting more than 24 hours is not an actual issue, but what i just realised is if a game goes past midnight, wht you just said will be true!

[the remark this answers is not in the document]
nocommentcode: Ok fixed it, actually we only need to store the time in miliseconds for the start time, everywhere else we can stick to hours and minutes since we aren't comparing it to anything

packages/TorneloScoresheet/src/components/TimePickerSheet/TimePickerSheet.tsx
```diff
@@ -0,0 +1,68 @@
...
+  gameTime,
+  visible,
+  dismiss,
+  setGameTime,
+}) => {
+  // set day/month/year not important
+  // date should be + 1 on hours for it to display properly,
+  // this is taken into account when setting the date
+  const [date, setDate] = useState(
+    new Date(1, 1, 1, gameTime.hours + 1, gameTime.minutes, 0),
```
gestureleft, line 26: I think it makes sense to make the default time the current time? Otherwise they may have to scroll through like 30 minutes everytime

gestureleft (changes requested):
Looks good! One small suggestion

packages/TorneloScoresheet/src/components/TimePickerSheet/TimePickerSheet.tsx
nocommentcode: Yes, that's what this code does, the `TimePickerSheet` component takes in the gameTime as an argument.
This gameTime is either:
- The current time saved on the Ply or
- The current time since the start of the game if there is no saved time on that Ply

The game time is being measured in hours and minutes.

gestureleft: Right, well I was just testing this and it came up with "1:00" even though the time was "1:38" or something, so there may be an issue somewhere? Does it work for you?

gestureleft: Hang on, I see what you mean, that means this move is one hour since the start of the game. Shouldn't it have been "0:01" or something then?

gestureleft: I'm not sure, but I think it would be clearer to have that UI be entering the time that the move happened? Might be one to check with David/Simon

nocommentcode: Yes the time is the time since the start of the game, i checked the rules.
So you log the time since the game started.
You should see 1 minute, did you see 1 hour?
Was the emulator left open for a long time?

gestureleft: Okay cool! Nice work.

Yep I saw one hour, the emulator had just been opened. In the code snippet this original comment was on it says:

```
new Date(1, 1, 1, gameTime.hours + 1, gameTime.minutes, 0)
```

Maybe it's the `+ 1` on the gameTime hours?

nocommentcode: Yeh so this must be an Ios, android compatibility thing,
On android, it wont display the time correctly if i dont have that +1 it shows me 24...
Since we're developing for ios, i guess ill change it to work there?

nocommentcode:
Ok i removed the +1 , does that work for you now @gestureleft

packages/TorneloScoresheet/src/components/TimePickerSheet/TimePickerSheet.tsx
gestureleft: Yeah nah, it crashes now. I'll try and diagnose

gestureleft (approved):
Okay I updated the podfile and fixed a typo, and now it should be good to go

PR #112, opened 2022-08-01
Graphics mode player card

B-Gunawan:
Added new Added `GraphicalModePlayerCard` component.
Might need a rename though.


B-Gunawan:
image

packages/TorneloScoresheet/src/style/colour.ts
```diff
@@ -13,10 +13,12 @@ export const colours = {
...
   secondary: '#1a2136' as const,
   secondary40: 'rgba(26, 33, 54, 0.4)' as const,
+  secondary70: 'rgba(26, 33, 54, 0.7)' as const,
   tertiary: '#ffbf00' as const,
   // Elements Colours
   darkenedElements: '#141414' as const,
   grey: '#6B6A6E' as const,
+  grey20: 'rgba(26, 33, 54, 0.2)' as const,
   darkGrey: '#323232' as const,
   // Chess Board Colours
```
gestureleft, line 23: This file no longer compiles since `statusBarStyleForColor` & `textColour` don't handle the new colours

packages/TorneloScoresheet/src/components/GraphicalModePlayerCard/GraphicalModePlayerCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const cardStyle = { ...styles.graphicalModePlayerCard, ...flexStyle };
+
+  const teamNameAndEloTextSection = (label: string) => (
+    <PrimaryText
+      size={15}
+      weight={FontWeight.Medium}
+      colour={colours.secondary70}
+      label={label}
+    />
+  );
```
gestureleft, line 27: Let's just make this a component? I can still be in this file, but it's a bit strange to make it a function when it is playing the role of a component

gestureleft (changes requested):
Couple of small things. Looks good though! Nice work

gestureleft:
Also - I think the grey is a bit darker than the figma? Maybe 10% grey would be better than 20%?

packages/TorneloScoresheet/src/style/colour.ts
B-Gunawan: Fixed

packages/TorneloScoresheet/src/components/GraphicalModePlayerCard/GraphicalModePlayerCard.tsx
B-Gunawan: Refactored to `CardTextRowComponent`

nocommentcode:
Ok I added the elo, team name and flags to this commit.
Also changed a few things:
- the background colour is now grey5 instead of 10 to match the figma
- I changed the text colours and sizes
- I clamped the name length to match the previous PR
- I added the vertical divider to match the figma
- I removed the components created since it's just wrapping the `PrimaryText` component for no reason

(note that the txt being cut off on the left hand side is a known bug with react native on android, apparently it shouldnt show up on ios)
image
image
image

nocommentcode (approved):
Nice, looks great

PR #113, opened 2022-08-02
Pin verification 

gwenllian1:
Done by Kat and Llio
- the app now checks if the pin is correct (it is currently hardcoded to be 1234)
- displays and error if  the pin is incorrect and autofocuses so the user can re-enter the input

https://user-images.githubusercontent.com/64511606/182361972-5fd0d802-4563-43e9-8151-55b965e23ce9.mp4

### DIFF
--- a/packages/TorneloScoresheet/src/components/Pin/Pin.tsx
+++ b/packages/TorneloScoresheet/src/components/Pin/Pin.tsx
@@ -18,9 +18,10 @@ const CELL_COUNT = 4;
 
 export type PinProps = {
   onPress: (pinCorrect: boolean) => void;
+  resetPin: number;
 };
 
-const Pin: React.FC<PinProps> = ({ onPress }) => {
+const Pin: React.FC<PinProps> = ({ onPress, resetPin }) => {
   const [value, setValue] = useState('');
   const ref = useBlurOnFulfill({ value, cellCount: CELL_COUNT });
   const [props, getCellOnLayoutHandler] = useClearByFocusCell({
@@ -33,6 +34,7 @@ const Pin: React.FC<PinProps> = ({ onPress }) => {
     <View>
       <SafeAreaView>
         <CodeField
+          key={resetPin}
           ref={ref}
           {...props}
           // Use `caretHidden={false}` when users can't paste a text value, because context menu doesn't appear
@@ -80,6 +82,7 @@ const Pin: React.FC<PinProps> = ({ onPress }) => {
             //incorrect pin
             setValue('');
             showError('Invalid Pin - Please Try Again');
+            onPress(false);
           }
         }}
         label="Enter Arbiter Mode"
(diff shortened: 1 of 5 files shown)


packages/TorneloScoresheet/src/components/Pin/Pin.tsx
```diff
@@ -9,11 +9,13 @@ import {
...
 import PrimaryButton from '../PrimaryButton/PrimaryButton';
 import PrimaryText from '../PrimaryText/PrimaryText';
+import { useError } from '../../context/ErrorContext';
+import { pinValid } from '../../util/arbiterPin';
 
 const CELL_COUNT = 4;
 
 export type PinProps = {
-  onPress: () => void;
+  onPress: (pinCorrect: boolean) => void;
```
nocommentcode, line 20: I think we should only be calling a function if the pin is correct right?
I might be missing something though... but you are never calling onPress(false) in the code
If we never need to let the page containing the pin know that the pin was incorect we should rename this to  onCorrectPinEntered()

packages/TorneloScoresheet/src/util/arbiterPin.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Given a pin string, return True if the pin is valid
+ */
+export const pinValid = (inputPin: string): boolean => {
+  const arbiterPin = '1234'; //TODO: change to global variable set by user.
+  return inputPin == arbiterPin ? true : false;
```
nocommentcode, line 6: === in js!
also no need for the ? true : false.
you can just leave it as `return inputPin === arbiterPin`

nocommentcode (changes requested):
Nice, looks good.
Just a tiny change regarding the function needed for the Pin component

packages/TorneloScoresheet/src/components/Pin/Pin.tsx
gwenllian1: We also call a function if the pin is false because it reloads the page

image
image

gwenllian1: so onPress we are calling handleVerify, and if the pin is valid it moves into arbiter mode but if it's incorrect it reloads the pin input so that the user can re-type

nocommentcode: hmm oh yeh I didn't see that, and is there any reason we are doing that?
Why is the reset pin comming from the parent component shouldn't the Pin component handle all the pin stuff by itself (ie the pinReset field should be a useState from within the Pin component)
Im bringing this up becuase the pin component is not only used ffrom the toolbar, its also used in the lock app thingy if the user switches out of the app and we might end up using it in other places too, so it would be better if the Pin compoment works all by itself with no need for extra inputs except the function to call once the pin intered is correct

gwenllian1: I hadn't realised it was used in other places sorry, I updated it so that everything happens in the pin component.

nocommentcode (approved):
perfect

PR #115, opened 2022-08-04
Enter player mode button

gwenllian1:
Kat and Llio
Added transitions from arbiter modes back to player modes

https://user-images.githubusercontent.com/64511606/182737152-fa0fd2b9-3e68-4cf4-a124-e1bfeda67a29.mp4

### DIFF
--- a/packages/TorneloScoresheet/src/components/Toolbar/Toolbar.tsx
+++ b/packages/TorneloScoresheet/src/components/Toolbar/Toolbar.tsx
@@ -2,6 +2,9 @@ import React, { useState } from 'react';
 import { Image, StatusBar, View } from 'react-native';
 import {
   useAppModeState,
+  useArbiterGraphicalRecordingState,
+  useArbiterResultDisplayState,
+  useArbiterTablePairingState,
   useGraphicalRecordingState,
   useResultDisplayState,
   useTablePairingState,
@@ -13,7 +16,11 @@ import {
   textColour,
 } from '../../style/colour';
 import { BLACK_LOGO_IMAGE, WHITE_LOGO_IMAGE } from '../../style/images';
-import { AppMode, isArbiterMode } from '../../types/AppModeState';
+import {
+  AppMode,
+  isArbiterFromPlayerMode,
+  isArbiterMode,
+} from '../../types/AppModeState';
 import IconButton from '../IconButton/IconButton';
 import PrimaryText, { FontWeight } from '../PrimaryText/PrimaryText';
 import Sheet from '../Sheet/Sheet';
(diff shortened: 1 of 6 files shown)


packages/TorneloScoresheet/src/hooks/appMode/arbiterGraphicalRecordingState.ts
```diff
@@ -0,0 +1,40 @@
...
+    const goToRecordingMode = (): void => {
+      const board = chessEngine.fenToBoardPositions(chessEngine.startingFen());
+      setAppModeState({
+        mode: AppMode.GraphicalRecording,
+        pairing: appModeState.pairing,
+        moveHistory: [],
+        board,
+        currentPlayer: appModeState.currentPlayer,
+      });
+    };
```
nocommentcode, line 37: This would reset the game to the starting state (empty move list and starting fen), we want to be able to resume the game...
So we should just copy over the current appModeState, the only thing that changes is the mode, which goes to its respective player mode
ie should be: 
`setAppModeState({ ...appModeState, mode: AppMode.GraphicalRecording });`

packages/TorneloScoresheet/src/hooks/appMode/arbiterResultDisplayState.ts
```diff
@@ -0,0 +1,36 @@
...
+    const [appModeState, setAppModeState] = useContext(context);
+    if (appModeState.mode !== AppMode.ArbiterResultDisplay) {
+      return null;
+    }
+
+    const goToResultDisplayMode = (): void => {
+      setAppModeState({
+        mode: AppMode.ResultDisplay,
+        pairing: appModeState.pairing,
+        result: appModeState.result,
```
nocommentcode, line 31: same deal here, we can just set it to `{...appModeState, mode: AppMode.ResultDisplay}`

packages/TorneloScoresheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -142,7 +142,7 @@ const Toolbar: React.FC = () => {
         )}
         {showPlayerModeButton && (
           <IconButton
-            icon="lock"
+            icon="lock-open"
```
nocommentcode, line 145: I'm thinking it might make more sense to invert these actually,
When you are in player mode, you want to 'unlock' the app into Arbiter mode.
When you're in Arbiter mode, you want to 'lock' the app back into player mode?
What do you think?

nocommentcode (changes requested):
Good stuff,
There's just an issue with the way we are transitioning back to the player mode in graphical recording (we're resetting the game)

packages/TorneloScoresheet/src/hooks/appMode/arbiterResultDisplayState.ts
nocommentcode: Though in this case and arbiterTablePairingState we are not changing anything, but writing it this way makes it so that if we change the state later on, we wont have to change this code here since the whole state will be copied over

nocommentcode (approved):
Perfect!

PR #116, opened 2022-08-04
fixed unecessary space that was annoying me

gwenllian1:
There was a space that was really annoying me so I removed it 
From this:
image
To this: 
image


ksov0001:
Nice :p Good pick up

ksov0001 (approved):
Looks good, good pick up :)

PR #118, opened 2022-08-04
Arbiter result display mode

dinz0014:
Arbiter Result Display Mode implemented by Dinu and Malo:

*Going back from Result Display to Recording Mode:*

[](https://user-images.githubusercontent.com/55499117/182766855-7a72cb28-6bf7-4f43-bc74-ee08d1eb83f3.mp4)

*Going back from Result Display to Enter Pgn:*

[](https://user-images.githubusercontent.com/55499117/182768558-a2fc99af-d5b4-444e-8881-e59a532b1324.mp4)

*Error state if there is an issue loading the Recording Mode data from storage:*

[](https://user-images.githubusercontent.com/55499117/182767199-fea65315-08c8-4e98-84cd-401e44101042.mp4)

### DIFF
--- a/packages/TorneloScoresheet/src/pages/Main.tsx
+++ b/packages/TorneloScoresheet/src/pages/Main.tsx
@@ -5,8 +5,8 @@ import { AppMode } from '../types/AppModeState';
 import EnterPgn from './EnterPgn/EnterPgn';
 import GraphicalRecording from './GraphicalRecording/GraphicalRecording';
 import PairingSelection from './PairingSelection/PairingSelection';
+import ArbiterResultDisplay from './ResultDisplay/ArbiterResultDisplay';
 import ResultDisplay from './ResultDisplay/ResultDisplay';
-import ArbiterTablePairing from './TablePairing/ArbiterTablePairing';
 import TablePairing from './TablePairing/TablePairing';
 
 const Main: React.FC = () => {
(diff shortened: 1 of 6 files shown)


ksov0001:
Small concern - when it enters Graphical Recording Mode from the Arbiter Result Page, isn't it better to go into the **Arbiter Graphical Recording Mode** instead? Because if the arbiter wants to continue making changes they would need to enter the pin twice otherwise. Alternatively, maybe the "Return to" popup can show an option for both?

packages/TorneloScoresheet/src/pages/Main.tsx
```diff
@@ -5,7 +5,9 @@ import { AppMode } from '../types/AppModeState';
 import EnterPgn from './EnterPgn/EnterPgn';
 import GraphicalRecording from './GraphicalRecording/GraphicalRecording';
 import PairingSelection from './PairingSelection/PairingSelection';
+import ArbiterResultDisplay from './ResultDisplay/ArbiterResultDisplay';
 import ResultDisplay from './ResultDisplay/ResultDisplay';
+import ArbiterTablePairing from './TablePairing/ArbiterTablePairing';
```
ksov0001, line 10: I think this file is missing in this branch. I can't see it on Git in the TablePairing folder in this branch. When I try to run this branch this line crashes it on my end. I can see this file in Arbiter_Table_Pairing_Branch though.

ksov0001 (approved):
Changes look good, nice work! 

Just a note for the future that it may need to return to Arbiter Graphical Recording Mode (instead of Player Graphical Recording Mode) but we can implement this in a later commit after agreeing on the best option.

nocommentcode (approved):
Lgtm

PR #119, opened 2022-08-04
Refactor the signature handling logic

gestureleft:
Written by Llio and I :)))

This PR was originally created to fix a bug, but we found some other small issues to fix along the way :)

- removed side effects in the ternaries
- fixed a bug where confirming the second player's signature did
nothing
- updated the styling of the result display in the signature component
- use a reference to store the state of the current player's signature;
clearly the signature library we are using is doing some crazy shit
with the callback we give it, because it ends up having stale state :(;
so instead of using state, we now use a reference, which never becomes
stale


nocommentcode (approved):
Noice

PR #120, opened 2022-08-05
Parse Team Name, Player Country and Elo from Tornelo PGN 

nocommentcode:
This PR updates our parsing logic to handle the changes made to the Tornelo API.
We now collect the Team Name, Player Country and Player Elo from the PGN.
This PR also updates the `PlayerCard` component by displaying this new information.
The PR also updates the styling of the `PlayerCard`, the `BoardPairing` and the `SignatureSheet` components to truncate large names as well as tidy up the styling of these.

**HINT: use this PGN from now on, it has the elos, team names and countries attached**:
https://staging-api.tornelo.com/api/divisions/ccfc669b-864f-4531-b5d9-99752ec6ba57/broadcast-pgn?convertMatchNumbers=false&includeDivisionName=false&token=3dbf8e4592fb80cba27a3afde4dc9f4d&withClocks=true

**New Pairing Selection style: (truncated names + fixed the board numbers)**

pairing selection new

**New TablePairing Page (new fields, truncated names and improved styling): **

table pairing new

**New Results Display Page (new fields, truncated names + updated styling):**

result display new

**New Confirm Winner signature (truncated names + updated styling): **
confirm result new


packages/TorneloScoresheet/src/components/PrimaryText/PrimaryText.tsx
```diff
@@ -36,10 +37,12 @@ const PrimaryText: React.FC<PrimaryTextProps> = ({
   style,
   size,
   children,
+  numberOfLines,
   align,
 }) => {
   return (
     <Text
+      numberOfLines={numberOfLines}
```
gwenllian1, line 45: Would it maybe make sense to default the number of lines to 1 if possible? Because I imagine that's what it will be in most places

gwenllian1 (changes requested):
This looks so much better :) I noticed a couple of minor things that could maybe  be changed a bit

gwenllian1:
I think this un-related to what you did, I suspect it might have happened when Harry and I were working yesterday but since you're making changes to the signature file I noticed that one of the results numbers is missing the colours.darkenedElements:
image
image

packages/TorneloScoresheet/src/components/PrimaryText/PrimaryText.tsx
nocommentcode: I don't think it would make sense to make it default, the number of lines argument truncates your text to fit on one line, I think if that is your target behaviour you should be setting it in the component, if it's default it might surprise us in some places

nocommentcode:
Ok i fixed the comments, What do you mean it doesnt want to recognise the module? Are you getting an error?

gwenllian1:
It looks so much better with the flags. I'm happy to approve, I just wanted to hear your thoughts about if a player has no country/flag because at the moment it looks like this: 
image
and I imagine as a player they might be confused because they won't know what is meant to be there, so I was wondering if it was worth maybe just not having anything there if the player doesn't have that information? But I'm aware that might not be worth our time to do right now.

nocommentcode:
> It looks so much better with the flags. I'm happy to approve, I just wanted to hear your thoughts about if a player has no country/flag because at the moment it looks like this: image and I imagine as a player they might be confused because they won't know what is meant to be there, so I was wondering if it was worth maybe just not having anything there if the player doesn't have that information? But I'm aware that might not be worth our time to do right now.

Yep agreed, looks a lot better now:
none res
none
all res
no flag
all

gwenllian1 (approved):

PR #121, opened 2022-08-10
Text recording mode setup

ksov0001:
**Kat + Llio**
- Created Text Recording Mode page.
- Refactored Recording Mode to not use two different states for Text Recording vs Graphical Recording (including impacted Arbiter Mode).
- Implemented a simple icon-based toggle in the toolbar (only visible in Recording mode) to quickly switch between Text Recording and Graphical Recording mode. 

Remaining in Future Iterations: 
- Implement animated toggle instead of icon for the switch (or potentially move out of toolbar)
- Implement functionality of text recording mode page

https://user-images.githubusercontent.com/41656704/183927952-34d777d6-ac40-48db-93d4-78fb6fb226d0.mp4


gwenllian1:
I wasn't sure what we should do for the tests because there are now two types of recording state,  will we need to do all of the test cases again in text recording mode?

packages/TorneloScoresheet/src/pages/Main.tsx
```diff
@@ -20,7 +20,12 @@ const Main: React.FC = () => {
     case AppMode.TablePairing:
       return <TablePairing />;
     case AppMode.Recording:
-      return <GraphicalRecording />;
+      return appMode.type === 'Recording' ? (
+        <GraphicalRecording />
+      ) : (
+        <PrimaryText label="Text Recording not currently supported" />
```
nocommentcode, line 26: Why not add the `<TextRecording />` page already implemented in this PR?

packages/TorneloScoresheet/__tests__/useAppModeState.test.ts
```diff
@@ -119,7 +119,7 @@ describe('graphical recording moving', () => {
 
     const graphicalState = generateGraphicalRecordingState([]);
```
nocommentcode, line 120: This test utility function needs to be renamed, and the variables should probably be changes to recordingState everywhere (you can easily do a find and replace) otherwise it might be confusing for people maintaining this code later on...

packages/TorneloScoresheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -43,36 +42,44 @@ const backgroundColorStyle = (backgroundColor: string) => ({
 
 const Toolbar: React.FC = () => {
   const appModeState = useAppModeState();
+  const graphicalRecordingState = useRecordingState();
```
nocommentcode, line 45: Probably makes sense to rename the variable here too

packages/TorneloScoresheet/src/pages/GraphicalRecording/GraphicalRecording.tsx
```diff
@@ -47,7 +47,7 @@ const otherPlayer = (player: PlayerColour | undefined) => {
 
 const GraphicalRecording: React.FC = () => {
   // app mode hook unpacking
-  const graphicalRecordingState = useGraphicalRecordingState();
+  const graphicalRecordingState = useRecordingState();
```
nocommentcode, line 50: This variable should probably be renamed as well, just for consistancy

nocommentcode (changes requested):
Looks good :) 
Just a few things:
- In the arbiterResultDisplayState,'s goToRecordingMode function, we need to set which type (probably default to Graphical)
- The toolbar component is getting pretty messy, it might make sense to transfer the arbiter mode / player mode button and this new toggle recording type button to their own components, so we can move the logic checking the mode away from the toolbar icon (we can do this in another PR)
- You need to merge Harry's PR into yours, he fixed some things but you have conflicts now
- No you dont need to write tests again, since the hook is the same!, we are only really testing the hook methods, so there are no extra tests to write, 
- However we do need to test the `toggleRecordingMode` function in the state, since it is a new function and is therefore untested
- There are still variables called graphicalRecording all over the place, do a global find to see all the occurences, the only thing still called GraphicalRecording should be the name of the page element.

nocommentcode:
> I wasn't sure what we should do for the tests because there are now two types of recording state, will we need to do all of the test cases again in text recording mode?

Actually there is only 1 type of recording state, recordingMode, and its custom hook is useRecordingState(), which is already tested intensively!

gwenllian1:
ok, I wasn't sure because I had to update the test_util to this:
image
which means we are only resting in the graphical type not in the text type at the moment so I wanted to double check we didn't also need to do the tests in text mode.

nocommentcode:
> ok, I wasn't sure because I had to update the test_util to this: image which means we are only resting in the graphical type not in the text type at the moment so I wanted to double check we didn't also need to do the tests in text mode.

Oh right i see, actually that field is just used to render either the Grphical recording page or the text recording page, it does not affect any methods in the hook (except for the toggleType() method as mentioned)

packages/TorneloScoresheet/src/hooks/appMode/recordingState.ts
```diff
@@ -155,7 +156,7 @@ export const makeUseGraphicalRecordingState =
 
     const goToArbiterGameMode = (): void => {
       setAppModeState({
```
nocommentcode, line 158: This call is missing the type, and we shouldn't be definining each property, i know this wasnt your code but it should be `{...appModeState, mode: AppMode.ArbiterRecording}` and that way we dont need to keep editing this every time we add a new field to recording mode

packages/TorneloScoresheet/src/hooks/appMode/arbiterRecordingState.ts
```diff
@@ -2,30 +2,34 @@ import { useContext } from 'react';
...
+    if (appModeState.mode !== AppMode.ArbiterRecording) {
       return null;
     }
 
     const goToRecordingMode = (): void => {
-      setAppModeState({ ...appModeState, mode: AppMode.GraphicalRecording });
+      setAppModeState({
+        ...appModeState,
+        mode: AppMode.Recording,
+        startTime: new Date().getTime(),
```
nocommentcode, line 31: This will reset the start time!
you can just leave this line out and it will copy the previous start time from the state

nocommentcode (changes requested):
Almost there just a couple issues

nocommentcode (approved):
perfect !!! :)))

PR #122, opened 2022-08-11
Editing moves first iteration

nocommentcode:
This PR implements the first iteration of move editing
In this PR, the user can now edit a move, if the move does not result in a broken state, the move is edited, otherwise, an error message is displayed.

This PR adds the following:
- an editMoveMode app mode
- a GraphicalEditMove page
- new unit tests

https://user-images.githubusercontent.com/26475724/184117738-f40da86f-9025-4791-ae15-8c6a77263936.mp4


packages/TorneloScoresheet/src/components/ChessBoard/ChessBoard.tsx
```diff
@@ -17,14 +17,10 @@ import { styles } from './style';
 export type ChessBoardProps = {
   positions: BoardPosition[];
   flipBoard?: boolean;
+  hilightedMove?: MoveSquares;
```
gestureleft, line 20: Typo

packages/TorneloScoresheet/src/pages/EditMove/GraphicalEditMove.tsx
```diff
@@ -0,0 +1,258 @@
...
+  // when the promotion popup opens, the app will await untill a promise is resolved
+  // this ref stores this resolve function (it will be called once the user selects a promotion)
+  const promotionSelectedFunc = useRef<
+    ((value: PieceType | PromiseLike<PieceType>) => void) | null
+  >(null);
+
+  // Button parameters
+  const actionButtons: ActionButtonProps[] = [
+    {
+      text: 'restart',
```
gestureleft, line 63: This actually wraps on an iPad because the word is too wide for the button. Maybe "reset" would be better? It conveys the same meaning but in less characters

packages/TorneloScoresheet/src/components/MoveCard/MoveCard.tsx
```diff
@@ -9,25 +9,40 @@ import { styles } from './style';
...
+  onRequestEditMove?: (colour: PlayerColour) => void;
 };
 
-const MoveCard: React.FC<MoveCardProps> = ({ move, onRequestEditMove }) => {
+const MoveCard: React.FC<MoveCardProps> = ({
+  move,
+  onRequestEditMove,
+  plyBeingEdited,
+}) => {
+  console.log(plyBeingEdited);
```
gestureleft, line 21: This is a little bit spammy

gestureleft (changes requested):
Love your work! Looks really good. I left some small comments.

`GraphicalEditMove` might be able to be simplified in a similar way to how #125 simplifies  `GraphicalRecordingMode`. But that can be saved for another PR, this is a great first pass at this

gestureleft (approved):
Brilliant

PR #123, opened 2022-08-12
simplified toolbar by putting arbiter/player mode transition and recording mode toggle in new files

gwenllian1:
In the last PR, Malo mentioned that the toolbar was getting cluttered so I separated out the toggle between text and graphical mode and the switching from arbiter to player mode.


nocommentcode (approved):
This is perfect! Thanks

PR #124, opened 2022-08-12
Add pgn comments

gwenllian1:
Add comments to the pgn for time recording and draw offers
image

### DIFF
--- a/packages/TorneloScoresheet/__tests__/chessEngine.test.ts
+++ b/packages/TorneloScoresheet/__tests__/chessEngine.test.ts
@@ -9,7 +9,6 @@ import {
   PlyTypes,
   ChessPly,
   GameTime,
-  MovePly,
 } from '../src/types/ChessMove';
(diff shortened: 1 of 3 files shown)


nocommentcode (changes requested):
Looks good, just a suggestion, we can simplify the comment thing quite a bit, 
Since we are playing the moves one by one, after making a move, we are sure there are no comments on that move yet right? 
So all we need to do at each move is make a comment with both the draw offer and the game time if either of them exist.
Also in js you can use the '`' characted to format a string, so we can do something like this:
`  game.setComment(`${move.drawOffer? "=" : "" }${move.gameTime? `[%clk ${move.gameTime.hours.toString()}:${move.gameTime.minutes.toString()}]` : ''}`)`

Also, we need to add test cases to test this functionality!, one for just a draw offer, one for just a game time and one for both

nocommentcode (approved):
Perfect!!

PR #125, opened 2022-08-14
Simplify GraphicalRecording component

gestureleft:
Basically just moved a bunch of stuff into it's own file, to make the
component file a bit easier to work with


nocommentcode (approved):
This is so much better

PR #126, opened 2022-08-15
[BUG] Fixed missing last move bug

nocommentcode:
When returning to recording mode from arbiter mode,
The last played move was not being displayed.
This was due to using the last move's starting fen to initialise the board,
Instead of computing the fen after the last move.


gestureleft (approved):
Perfection

gwenllian1 (approved):
looks good :)

PR #127, opened 2022-08-18
Email Game Results

ksov0001:
Implemented Email Game button on Results page. Button opens default email client with body and subject of email prefilled to contain game info and PGN. User inputs the desired email address in the client and then after sending the email is returned to the Results page. 

**Important notes** 
- App locking on exit has been disabled on the Results page for this feature to be usable (so that arbiter isn't required to input pin every time someone sends an email). 
- This feature doesn't work on the Android emulator - for testing please run on a real device. 
- This feature cannot support attachments or embedded images, so the MVP email does not contain the signatures of the players. This can be resolved in the future by transitioning to Tornelo email API instead. 

**Demo Video** 
Note - since it had to be run on a real device the demo looks really bad since I'm using an Android phone! I needed to relocate the button to the top of the screen for the demo only - in the actual version the button has not been moved.

https://user-images.githubusercontent.com/41656704/185310502-555a7edc-8619-4f4a-be80-9fe947fdc847.mp4


packages/TorneloScoresheet/src/pages/ResultDisplay/ResultDisplay.tsx
```diff
@@ -21,8 +22,25 @@ const ResultDisplay: React.FC = () => {
   }`;
 
   const handleEmailGame = (): void => {
-    // TODO: Implement email logic
-    console.log('email game button pressed');
+    const subject =
```
gwenllian1, line 25: Could this also be in the send email file?  That way it's all in the same place and it keeps the result display file purely as the result display page.

gwenllian1:
I think it would be a good idea to confirm that this works for Harry on the iPad before we merge into main.

packages/TorneloScoresheet/src/pages/ResultDisplay/ResultDisplay.tsx
```diff
@@ -21,8 +22,25 @@ const ResultDisplay: React.FC = () => {
...
+      'Tornello Game Results ' +
+      infoString +
+      ':' +
+      resultDisplayMode?.pairing.players[0].firstName +
```
gwenllian1, line 29: The pgn already has thing information so since it's in the subject we might not need to include it here? It doesn't really matter though
image

packages/TorneloScoresheet/src/pages/ResultDisplay/ResultDisplay.tsx
```diff
@@ -21,8 +22,25 @@ const ResultDisplay: React.FC = () => {
...
+      ' vs' +
+      resultDisplayMode?.pairing.players[1].firstName;
+    const body =
+      'Tornello Game Results ' +
+      infoString +
+      '\n\nPlayer 1: ' +
+      resultDisplayMode?.pairing.players[0].firstName +
+      resultDisplayMode?.pairing.players[0].lastName +
```
gwenllian1, line 37: We might not need to include the player names here since it's already in both the subject of the email and the pgn. But if you want to include it we have a function in player  (in utils) called fullName which returns the full name of the player so we should probably use that instead to shorten the code.

packages/TorneloScoresheet/src/pages/ResultDisplay/ResultDisplay.tsx
nocommentcode: Agreed, the email function should probably take the email, `ChessGameResult`, and  `ChessGameInfo` as arguments

packages/TorneloScoresheet/src/pages/ResultDisplay/SendEmail.tsx
```diff
@@ -0,0 +1,24 @@
...
+  }
+  // check if we can use this link
+  const canOpen = await Linking.canOpenURL(url);
+
+  if (!canOpen) {
+    throw new Error('Provided URL can not be handled');
+  }
+
+  return Linking.openURL(url);
+}
```
nocommentcode, line 24: I feel like this should probably be a function that lives in the `util` directory,
maybe in a new file called `emailUtils` ?
I mean we can leave it here but its better to leave only UI stuff in the pages directory

packages/TorneloScoresheet/src/pages/ResultDisplay/ResultDisplay.tsx
```diff
@@ -22,15 +22,24 @@ const ResultDisplay: React.FC = () => {
...
+      '\nPlayer 2: ' +
+      resultDisplayMode?.pairing.players[1].firstName +
+      resultDisplayMode?.pairing.players[1].lastName +
+      '\n\nFull Game PGN:\n' +
+      resultDisplayMode?.result.gamePgn;
     sendEmail('', subject, body); //option to add arbiter 'To' address in the future
```
nocommentcode, line 43: We aren't setting the email here?
We need to have a popup sheet come up when we press 'email game' that asks the user to input an email address.
Once the user clicks send on that popup, then we send the email with that email address

packages/TorneloScoresheet/src/pages/ResultDisplay/SendEmail.tsx
```diff
@@ -0,0 +1,24 @@
...
+    url += `?${query}`;
+  }
+  // check if we can use this link
+  const canOpen = await Linking.canOpenURL(url);
+
+  if (!canOpen) {
+    throw new Error('Provided URL can not be handled');
+  }
+
+  return Linking.openURL(url);
```
nocommentcode, line 23: I think this function should return a Promise<Result<string>> here, because the linking.openURL will throw an exception if there is a problem, so we need to catch that exception and return a failure so that we can display the error toast

nocommentcode (changes requested):
Good stuff!
Just a few things, 
We are missing the popup for the user to input an email address in, 
It might be best to create a new `EmailResults` Component that will handle this popup and on submitting an email, it calles the `sendEmail` function.
It should probably also show an Error toast if the sendEmail returns a failure

packages/TorneloScoresheet/src/pages/ResultDisplay/SendEmail.tsx
nocommentcode: Oh wait sorry i just saw the line above it, in our project, we are avoiding throwing exceptions, instead we return a Result object, which can be handled by the client calling this function to display an error

nocommentcode:
Also I found this tutorial on how to add attachments : https://jaishil-bhavsar.medium.com/react-native-attachment-in-email-by-moving-file-from-internal-storage-to-external-60b9e0b97da7
Seems like it is possible, just a bit harder

packages/TorneloScoresheet/src/pages/ResultDisplay/ResultDisplay.tsx
ksov0001: I've changed it to use the fullName function - I think I'll keep the names in the email body since they are a bit harder to find the names in the PGN.

packages/TorneloScoresheet/src/pages/ResultDisplay/ResultDisplay.tsx
ksov0001: Sure, I've added this now. I must have misunderstood the requirement since in the Trello it said "The players are obliged to submit the text of their game to the Organizer with reference to article 8.3 of the Laws of Chess. For the MVP, it is sufficient to have a button that emails the file to Tornelo" so I assumed it would just send to one email address. I've just made it so that it asks the user to input their desired email address :)

ksov0001:
Thanks Llio and Malo! I have made the recommended changes:
- Refactored to show error message / using promises /Result return instead of throwing error
- Moved function to util and changed so that the email util handles building the email text
- Added email input option in the app 

See updated demo videos below. 

https://user-images.githubusercontent.com/41656704/185549625-19290f49-4735-414b-9ca0-ced934bc6a26.mp4

https://user-images.githubusercontent.com/41656704/185549743-23097781-d16f-4299-b15e-eae22d12c4dd.mp4

The only change I'm not making in this PR is a spike into the attachments - let's investigate this in a separate PR :)

nocommentcode (approved):
Perfect!

nocommentcode (approved):
good stuff

PR #129, opened 2022-08-22
Add the application version to the toolbar

gestureleft:


nocommentcode (approved):
Looks good, only thing is we need to remove this for a production version, we can do this later, but perhaps in package.json we need a production flag or something so we know to hide the version

gestureleft:
@nocommentcode Yep good idea- I think maybe it should be an env var though? That way, in CI in future, we can just pass a flag to say whether it's prod or not and choose to render the version based on that

PR #132, opened 2022-08-30
Implement User documentation

gestureleft:
This patch implements user documentation in the toolbar.

When the question mark is pressed in the toolbar, a modal is rendered to the screen with documentation for the current page.

The documentation is stored in a series of markdown files that are bundled at build time and then dynamically loaded from the file system at run time.

This feature is working on iOS, but there may be some changes required to get it working on Android (specifically: updating android config to bundle the `assets/documentation` directory at build time)

TorneloScoresheet help rendering


nocommentcode (approved):
lgtm

PR #134, opened 2022-08-31
Updated iOS configuration for deployment

gestureleft:
In order to deploy the app for testflight, we had to make some changes to the iOS configuration.


nocommentcode (approved):
lgtm

PR #140, opened 2022-09-17
Implement undo & redo stacks in Graphical Recording Mode

gestureleft:
Certain actions can now be undone and redone. The actions are maintained on a "stacks" that the UI logic can push to and pop from

Written by Harry + Dinu

### DIFF
--- a/packages/TorneloScoresheet/src/util/moveHistory.ts
+++ b/packages/TorneloScoresheet/src/util/moveHistory.ts
@@ -1,0 +1,35 @@
+import { chessEngine } from '../chessEngine/chessEngineInterface';
+import { ChessPly, PlyTypes } from '../types/ChessMove';
+
+/**
+ * Gets the current fen of the game based on move history
+ * @param moveHistory array of ChessPly
+ * @returns current fen
+ */
+export const getCurrentFen = (moveHistory: ChessPly[]): string => {
+  // no moves -> starting fen
+  if (moveHistory.length === 0) {
+    return chessEngine.startingFen();
+  }
+
+  // execute last ply to get resulting fen
+  const lastPly = moveHistory[moveHistory.length - 1]!;
+
+  // Last ply = SkipPly
+  if (lastPly.type === PlyTypes.SkipPly) {
+    return chessEngine.skipTurn(lastPly.startingFen);
+  }
+
+  // Last ply = MovePly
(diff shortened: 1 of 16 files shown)


packages/TorneloScoresheet/src/util/moveHistory.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  // Last ply = MovePly
+  return (
+    chessEngine.makeMove(
+      lastPly.startingFen,
+      lastPly.move,
+      lastPly.promotion,
+    ) ?? '' // all move in history are legal, -> should never return undef
+  );
+};
```
nocommentcode, line 31: Should you remove this function from the `recordingState` file then and use the util version from the recording hook?

nocommentcode (changes requested):
There are comipilation errors in enterPgn page, where you changed the state but not the page,
Its crashing when trying to undo, getting a 'stack.at is undefined' error
Also it is significantly slower now, it takes like 2 of 3 seconds to process a move i dont think this is acceptable
Also the are we not going to support undoing editing moves?
Also there are no unit tests for the undo stack hook

packages/TorneloScoresheet/src/util/moveHistory.ts
gestureleft: This has been removed from the `recordingState` file

gestureleft:
> There are comipilation errors in enterPgn page, where you changed the state but not the page,

Compilation errors fixed.

> Its crashing when trying to undo, getting a 'stack.at is undefined' error

This is strange - `stack.at` is definitely [a thing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) and is working on my machine. I guess it's possible that the Android version of Hermes doesn't support it yet for some reason?

> Also it is significantly slower now, it takes like 2 of 3 seconds to process a move i dont think this is acceptable

The performance is identical on my machine. I don't see why there would be any change - the only extra overhead is pushing to an array.

> Also the are we not going to support undoing editing moves?

Yep - this PR is already big enough though, so that'll be added in a future PR :^)

> Also there are no unit tests for the undo stack hook

Good point - I'll get on this now!

nocommentcode (approved):
Its working for me now,
Its still considerable slower now though...

PR #141, opened 2022-09-17
Slight tweaks to the UI for the View Past Games

gestureleft:
Now the way to access the past games is in the toolbar. And the way to exit is also in the toolbar


nocommentcode (approved):
good stuff

nocommentcode (approved):
!

PR #142, opened 2022-09-18
Legality checking (Updated)

nocommentcode:
This PR implements @ksov0001  adn @gwenllian1  's previous pr but i slightly changed the way we are doing legality checking to make it easier to implement my ticket
also changed the chess engine interface to avoid performing moves twice when getting the san and fen of a move
Also fixed the edit move state to change the occurence count after editing a move (this was a bug)
Added extra unit test cases to reflect these changes


ksov0001 (approved):
Reviewed with Llio - looks good

PR #143, opened 2022-09-18
Editing add skips

gwenllian1:
If a move is edited, all future moves that are made illegal are replaced with skips
Kat + Llio.

https://user-images.githubusercontent.com/64511606/190897721-dab5e10d-c5a7-4138-af3c-c5c0c2d959be.mp4

### DIFF
--- a/packages/TorneloScoresheet/src/hooks/appMode/editMoveState.ts
+++ b/packages/TorneloScoresheet/src/hooks/appMode/editMoveState.ts
@@ -93,15 +93,22 @@ const rebuildHistory = (
   }
   let previousPly: ChessPly = editedPly;
   let previousStartingFen = getStartingFen(previousPly) ?? '';
-  if (previousStartingFen === '') {
-    return null;
-  }
 
   // for each move after the edited ply
   // we generate it's new starting fen based on the previous ply
   // if the starting fen is null (the move was impossible) we add null to to newHistory[]
   // otherwise we add the ply with the updated starting fen to newHistory[]
   const newHistory = futureMoves.map(move => {
+    if (previousStartingFen === '') {
+      previousPly = {
+        moveNo: move.moveNo,
+        startingFen: previousPly.startingFen,
+        drawOffer: move.drawOffer,
+        player: move.player,
+        type: PlyTypes.SkipPly,
+      };
+      return previousPly;
+    }
     positionOccurances = updatePositionOccurence(
       positionOccurances,
       previousStartingFen,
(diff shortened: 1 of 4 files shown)


packages/TorneloScoresheet/src/hooks/appMode/editMoveState.ts
```diff
@@ -96,15 +97,22 @@ const rebuildHistory = (
...
   // if the starting fen is null (the move was impossible) we add null to to newHistory[]
   // otherwise we add the ply with the updated starting fen to newHistory[]
   const newHistory = futureMoves.map(move => {
+    if (previousStartingFen === '') {
+      previousPly = {
+        moveNo: move.moveNo,
+        startingFen: previousPly.startingFen,
+        drawOffer: move.drawOffer,
+        player: move.player,
+        type: PlyTypes.SkipPly,
```
nocommentcode, line 112: Missing the legality check here, it should be performed with the fen after this skip ply is recorded

packages/TorneloScoresheet/src/hooks/appMode/editMoveState.ts
```diff
@@ -96,15 +97,22 @@ const rebuildHistory = (
...
+      };
+      return previousPly;
```
nocommentcode, line 114: Also we need to update the posistion occureneces with this new fen before returning previous ply

packages/TorneloScoresheet/src/hooks/appMode/editMoveState.ts
```diff
@@ -118,8 +126,21 @@ const rebuildHistory = (
...
+      // return a skip if move impossible
       if (!nextMoveResult) {
-        return null;
+        previousPly = {
+          moveNo: move.moveNo,
+          startingFen: previousStartingFen,
+          drawOffer: move.drawOffer,
+          player: move.player,
+          type: PlyTypes.SkipPly,
+          legality: checkMoveLegality(previousStartingFen, positionOccurances),
```
nocommentcode, line 138: The legality of this move needs to be checked with the fen resulting from this move, ie you need `const nextFen = chessEngine.skipTurn(previousStartingFen)` before declaring previousPly, then you need to call checkMoveLegality with this next fen,
 I just noticed there is a bug in the previous version, in line 138, it should be checking the legality with the next fen too!

packages/TorneloScoresheet/src/hooks/appMode/editMoveState.ts
nocommentcode: and then we need to set the previous starting fen to be this new fen

nocommentcode (changes requested):
Looks good, just needs a minor fix with the legality checking.
I also realised there was a bug with the previous version, I was checking the legality with the previous fen instead of the next fen for skip plys, so this needs to be changed!

nocommentcode (approved):
nicely done :)

nocommentcode (approved):
!

ksov0001 (approved):
Re-approving after rebase

PR #144, opened 2022-09-19
Arbiter mode display legality

nocommentcode:
This PR implements highlighting the moves in arbiter mode to represent legality.
For now all repetition moves are highlighted the same colour, in the future, it might make more sense to highlight different repetitions in different colours so the arbiter can make more sense of it.

The help sheet explains what the colours mean, but it is not working on android

**In this example there is repetition and a checkmate:**

https://user-images.githubusercontent.com/26475724/190940811-51b2bf19-e0a9-4f5c-b531-b3d1de0e50ca.mp4

**Stalemate:**

https://user-images.githubusercontent.com/26475724/190940818-f454bf06-3210-4bf5-a644-3fb06b922881.mp4

### DIFF
--- a/packages/TorneloScoresheet/assets/documentation/ArbiterRecording.md
+++ b/packages/TorneloScoresheet/assets/documentation/ArbiterRecording.md
@@ -1,0 +1,10 @@
+# Arbiter Recording
+
+This page allows the arbiter to view all the past moves as well as their legality.
+
+The moves are visible in the table and are colour coded to represent differing legality statuses:
+
+- A Blue move means that move is in repetition, all other instances of that repetition are also in Blue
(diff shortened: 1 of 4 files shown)


gestureleft (changes requested):
Looks great!

One small thing would be that when scrolling up above the top of the list, the black border stays still while the list scrolls. It's actually hard to explain so I've attached a clip: 

https://user-images.githubusercontent.com/57308619/190974723-473ce59c-c809-422b-9ad9-87415685ba0b.mov

The styling also looks slightly weird when there are no moves (there is a black line) - maybe remove the line and render a message saying there are no moves?

Nice work!

gwenllian1 (approved):

PR #147, opened 2022-09-25
Support ending the game with skip plys

nocommentcode:
The user can now end the game even if there are skip plys recorded
This involved changing the chess.ts engine to support skip moves in order to be able to generate a pgn with them.
Updated unit tests in chess.ts

**Note: due to changes in chess.ts, you will have to make sure you recompile chess.ts using `npx tsc`**
Added a new argument to chess engine to determin if game should be allowed to end with skips

Edited front end to support this change:

https://user-images.githubusercontent.com/26475724/192129408-59395869-18e3-49ee-9e26-95ce6379b96a.mp4


gwenllian1:
incredibly super minor thing but the message 'You still have skips recorded in the game, are you sure you want to end the game with skips recorded?" is maybe a bit too long. I think if you shorten it to something like 'are you sure you want to end the game with skips?' would look a bit nicer on the UI

nocommentcode:
I changed the message @gwenllian1

gwenllian1 (approved):

PR #150, opened 2022-10-05
Simplified the GraphicalRecording ViewModel

gestureleft:
Rather than having a seperate function for moving with skips and just moving - we handle that internally now

The idea here is to simplify the GraphicalRecording View's logic and reduce code duplication


nocommentcode (approved):
much better

PR #152, opened 2022-10-06
Click to Move Kat + Llio

gwenllian1:
- Implemented click to move in recording and editing modes
- We noticed a lot of similarities between the recording mode and editing mode state code so it might be worth having a single file for the functionality they both have.

https://user-images.githubusercontent.com/64511606/194217832-6721e771-92d2-49f0-b2be-6db53820878b.mp4

https://user-images.githubusercontent.com/64511606/194217839-d7b1f5d0-e4f2-465a-a524-976c629a6650.mp4


packages/TorneloScoresheet/src/components/DragAndDrop/DragAndDropContext/DragAndDropContext.tsx
```diff
@@ -87,6 +90,67 @@ export const useHitTest = (
...
+};
+const ClickToMoveContext = React.createContext<
+  [
+    ClickToMoveContextState,
+    React.Dispatch<React.SetStateAction<ClickToMoveContextState>>,
+  ]
+>([{ fromSquare: null, toSquare: null }, () => undefined]);
+
+type clickToMoveHook = {
+  registerSquare: (postiion: Position | null) => void;
```
nocommentcode, line 105: typo

packages/TorneloScoresheet/src/components/DragAndDrop/DragAndDropContext/DragAndDropContext.tsx
```diff
@@ -87,6 +90,67 @@ export const useHitTest = (
...
 
+type ClickToMoveContextState = {
+  fromSquare: unknown | null;
+  toSquare: unknown | null;
+};
+const ClickToMoveContext = React.createContext<
+  [
+    ClickToMoveContextState,
+    React.Dispatch<React.SetStateAction<ClickToMoveContextState>>,
+  ]
```
nocommentcode, line 101: also all this clikc to move stuff should be in its own file instead of drag and drop context's file.

packages/TorneloScoresheet/src/hooks/appMode/recordingState.ts
```diff
@@ -359,21 +398,58 @@ export const makeUseRecordingState =
...
+      }
+      move(
+        {
+          from: pressToMoveSelectedFromSquare.position,
+          to: position,
+        },
+        promotion,
+      );
+      setPressToMoveSelectedFromSquare(undefined);
+    };
```
nocommentcode, line 433: It also looks like you're not using the context you created before? Instead we are putting this logic in the recording state,
Wouldn't it be better to store this logic in its own context? 
I think it would be better for 2 reasons:
- So that we do not make the recording state's larger than it should be
- This click to move logic is a front end consideration, whereas the recording state should be storing business logic related to the view model (the recording state)
- It means you can resuse this functionality wiithout having to copy code (for example in the edit move state you've had to repeat all these functions, if they are stored in a separate context, you wouldnt have to, you'd only need to pass through the move function into the context)

nocommentcode (changes requested):
looking good :)
However I think we should move this click to move logic into a separate context, it will be much cleaner that way, 
You actually already have the context coded in a file but you're not using it!
It would work similar to drag and drop context, the ChessBoard component already has a   onMove: (moveSquares: MoveSquares) => Promise<void> property, you can tap into that function and it would require 0 cahnges to the state, you just need the click to move context to call that function once it has regitered a to and from postion, all the pawn promotion stuff and what not will already be handled by this onMove function which is already coded for both the recording and edit move pages, so there's no need to redo all that logic

nocommentcode (approved):
looks good :)

PR #153, opened 2022-10-10
Updated the default value for game time

gestureleft:
We now use the game time of the most recent move with a game time as the default value for editing the time of a move. If there is no such game time, we use zero


ksov0001 (approved):
Looks great :)

PR #154, opened 2022-10-10
Fixed bug where move times were formatted poorly in pgn

gestureleft:
We weren't adding a leading zero to minute and hour values less than 10


ksov0001 (approved):
Nice work

PR #155, opened 2022-10-10
Added headings to game time picker

gestureleft:


ksov0001 (approved):
Nice work! Agreed that maybe in the future the layout may be improved (eg. making the number scroll a bit larger) but for MVP this is a great improvement.

PR #157, opened 2022-10-10
Do nothing if player drops piece in it's origin square

gestureleft:
Previously we would try and make a move, which obviously wouldn't work, since you can't move a piece from and to the same square in chess.

This fixes a bug where we show an error message if the user changes their mind about moving a piece


ksov0001 (approved):
Nice to have that fixed! Looks good :)

PR #163, opened 2022-10-14
Highlight repeated move cells in Arbiter table (if final move is repeated) 

ksov0001:
Highlights last cell in Arbiter recording mode table if it's in 3+ repetition, plus all other moves that are of the same fen after the move. 

https://user-images.githubusercontent.com/41656704/195847150-0e4b9bef-9d9a-4cc0-8bec-88ac85595a37.mp4


nocommentcode (approved):
looks good to me

PR #164, opened 2022-10-15
Arbiter needs a way to load a different game without entering a result

gwenllian1:
Previously, if an arbiter wanted to select a different game they would have to enter a game result and go back from result mode. This allows them to return to pairing selection directly from recording mode. 

https://user-images.githubusercontent.com/64511606/195967565-50053a18-b4e2-4367-84b0-e36bbeb03acb.mp4


ksov0001 (approved):
Nice, looks good! Very useful feature

nocommentcode (approved):
run pipeline

PR #165, opened 2022-10-15
Fix fen position occurence counting

nocommentcode:
There was an issue with the way we were counting position occurences.
Basically, a fen is composed of multiple sections (https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation)
We were splitting the fen by the '-' character, but this would break if an en passant square exists, then the '-' would not be in the fen.
This PR fixes this problem by introducing a fen util function called `getStateFromFen` which returns just the state of the board and whos turn it is from the fen, which is the only information we need to be counting position occurences


ksov0001 (approved):
Good pick up, this is definitely much cleaner than before. Looks good :)

PR #166, opened 2022-10-16
Standardise Arbiter Mode navigation

nocommentcode:
This PR standardises the way the arbiter will navigation between app modes.
Previously some arbiter modes had a button to select which mode to return to, some arbiter modes had a back button, and some arbiter modes used a back button in the toolbar.
This PR makes it so all arbiter modes use a hamburger menu icon in the toolbar to select which mode to go to.
This opens an option sheet that the arbiter can use to select which app mode they want to move to 

This PR also removed the transition functions from the toolbar hook and moved them to the respective app mode hooks, which makes stuff much cleaner by avoiding the if else statements in the toolbar state to find out which app mode we are in 

https://user-images.githubusercontent.com/26475724/196017736-44c6290c-7cec-4d6b-bb1d-b292b47b59de.mp4

### DIFF
--- a/packages/TorneloScoresheet/src/components/Toolbar/ArbiterModeNavigation.tsx
+++ b/packages/TorneloScoresheet/src/components/Toolbar/ArbiterModeNavigation.tsx
@@ -1,0 +1,137 @@
+import React, { useState } from 'react';
+import { View } from 'react-native';
+import {
+  useAppModeState,
+  useArbiterRecordingState,
+  useArbiterResultDisplayState,
+  useArbiterTablePairingState,
+  useEnterPgnState,
+  usePairingSelectionState,
+  useViewPastGames,
+} from '../../context/AppModeStateContext';
+import { AppMode } from '../../types/AppModeState';
+import IconButton from '../IconButton/IconButton';
+import OptionSheet, { Option } from '../OptionSheet/OptionSheet';
+import { styles } from './style';
+
+const ArbiterModeNavigation: React.FC = () => {
+  const appModeState = useAppModeState();
+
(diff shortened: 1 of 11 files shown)


packages/TorneloScoresheet/src/components/Toolbar/ArbiterModeNavigation.tsx
```diff
@@ -0,0 +1,139 @@
...
+    [AppMode.ArbiterTablePairing]: true,
+    [AppMode.ArbiterResultDisplay]: true,
+    [AppMode.PairingSelection]: true,
+    [AppMode.ViewPastGames]: true,
+
+    // do not show arbiter navigation button
+    [AppMode.ResultDisplay]: false,
+    [AppMode.TablePairing]: false,
+    [AppMode.EditMove]: false,
+    [AppMode.EnterPgn]: true,
```
gwenllian1, line 33: this is under do not show arbiter navigation button but is set to true

gwenllian1 (changes requested):
I really like this, I think it's much better than what we had before.

gestureleft:
Just wanted to chime in on the design! I really like this idea, but not sure if a hamburger menu icon is the right icon to go with, since it doesn't open a hamburger menu? So I reckon we either use a different icon, or put the navigation in a hamburger menu. New icon would probably be easiest. Or we could leave it as is, and leave implementing a hamburger menu as an exercise for the reader 🤔

nocommentcode:
yep agreed, I changed it to an arrow down icon

gwenllian1 (approved):

PR #170, opened 2022-10-18
Implemented undoing and redoing editing a move

gestureleft:
This is a massive patch that changes a lot of files. Sorry about that. This change kind of necessitated it.

Summary:
 - Reworked tests to not assume that a value is passed to `useState` state setters (we are now passing functions in a bunch of places)
 - Consolidated move editing and move recording into one component
 - Rewrote a bunch of state logic to support being called multiple times between renders (this also revealed a couple of bugs relating to position occurences, and this patch fixes those)
 - Implemented undoing editing moves in the past


nocommentcode (changes requested):
damn gg bro looking good!
The code all looks good to me,
I played around with it a bunch, the editing is working fine, 
The moving is working fine
But it did notice a few issues with the redo functionality, not too sure what the deal is but basically if you undo and edit move and then redo it the behaviour is not what you'd expect
Ive attached a few examples of this:

https://user-images.githubusercontent.com/26475724/196565653-a5056c77-963b-41fa-841b-0cda46c29042.mp4

https://user-images.githubusercontent.com/26475724/196565645-d31777d8-b949-4cea-82a4-99a735aa209f.mp4

Also the undo function can beahve weiredly with skips:


https://user-images.githubusercontent.com/26475724/196565810-b38d2592-2b46-4ebe-8287-f366f44e9f8e.mp4

nocommentcode:
superceeded by #174

PR #171, opened 2022-10-19
Add classic chess pieces

nocommentcode:
This PR adds the Classical Chess Pieces provided by David.
THis PR Also edits the promotion sheet and Chess board components to accept arguments as to which style of piece they should render.
This defaults to tornelo style.


B-Gunawan (approved):
LGTM. Could probably add a global alias for selecting piece styles. but probably not worth the effort rn.

nocommentcode:
I think the PieceAsset component will handle deciding which piece to render, ill change the other pr so that the piece asset uses the general settings hook so that there is no need for extra arguments

PR #172, opened 2022-10-19
General settings

nocommentcode:
This PR creates a genera settings context, as well as a settings sheet which can be accessed from the toolbar.
For now the only setting is the chess peice style

**Chess Board:**

https://user-images.githubusercontent.com/26475724/196586156-2f644624-d9ce-4fc8-a6fd-269c8012963b.mp4

**Promotion sheet: **

https://user-images.githubusercontent.com/26475724/196584499-b539e1eb-4723-4710-8a79-34bc7cc6b224.mp4

### DIFF
--- a/packages/TorneloScoresheet/App.tsx
+++ b/packages/TorneloScoresheet/App.tsx
@@ -12,6 +12,7 @@ import { GestureHandlerRootView } from 'react-native-gesture-handler';
 import LockAppOnExit from './src/components/LockAppOnExit/LockAppOnExit';
 import { ModalStackContextProvider } from './src/context/ModalStackContext';
 import SplashScreen from 'react-native-splash-screen';
+import { GeneralSettingsContextProvider } from './src/context/GeneralSettingsContext';
 
 const App = () => {
   const isDarkMode = useColorScheme() === 'dark';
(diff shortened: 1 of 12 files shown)


gwenllian1:
Looks good, my only suggestion would be that it feels kind of weird to change the settings and not click anything to confirm that so maybe we could add a confirm button and pressing the X to close just closes the popup without making a change? that way we are sticking with the same confirm and cancel pattern that we are using throughout the app :) 

I'm also getting this warning now (I don't know if that's to do with anything you changed but thought I'd mention it) 
image

gwenllian1:
Also I didn't change any settings but it defaulted to the 'classic' style but in your other PR you mentioned it should default to the Tornelo style

B-Gunawan:
Following up on comment on #171: maybe to reduce checks on the promotions sheet, could store the piece assets in the general setting context? It might help in abstracting the piece style selection from individual components.

nocommentcode:
@gwenllian1 yes sorry i mean to say classical style, its an easy fix to change if they prefer it in tornelo style, so for now it defaults to classical style.
@B-Gunawan I've abrstracted everything behind the pieceasset component, so it now decides which piece to render based off the genreal setting
@gwenllian1 that warning is linked to the qr code library we are using, it is because one of its dependencies is old and depreciated, it doesnt cause any errors though

gwenllian1 (approved):

PR #174, opened 2022-10-21
Implemented undoing and redoing editing a move

nocommentcode:
This is my fix of harrys PR,
I fixed a few issues, there are still some quirky behaviours at time, 
This will need to be looked into,
But we dont have time to fix this now, so lets jsut merge it in its working almost perfectly theres just some fringe cases

This is a massive patch that changes a lot of files. Sorry about that. This change kind of necessitated it.

Summary:
 - Reworked tests to not assume that a value is passed to `useState` state setters (we are now passing functions in a bunch of places)
 - Consolidated move editing and move recording into one component
 - Rewrote a bunch of state logic to support being called multiple times between renders (this also revealed a couple of bugs relating to position occurences, and this patch fixes those)
 - Implemented undoing editing moves in the past


gestureleft (approved):
lgtm 🔥 🚒

PR #176, opened 2022-10-25
Fix general settings loading bug

nocommentcode:
There was a bug with the general settings where the settings were not saved after closing the app.

Actually the settings were being saved, but were overriden by the settings sheet at the start due to calling useEffect() for both the settings context and the settings sheet (the sheet would finish first since it is not async, and would then ovveride the saved settings).
To avoid this problem, we will only store the settings after pressing save


gestureleft (approved):
LGTM

PR #178, opened 2022-10-25
Bumped version to 1.4

gestureleft:


packages/TorneloScoresheet/ios/Podfile
```diff
@@ -13,13 +13,10 @@ target 'TorneloScoresheet' do
   )
 
 
-  pod 'react-native-camera', :path => '../node_modules/react-native-camera'
-
-  pod 'RNPermissions', :path => '../node_modules/react-native-permissions'
-
   permissions_path = '../node_modules/react-native-permissions/ios'
```
nocommentcode, line 16: are you sure this isnt needed for the qr code scnanung stuff?

nocommentcode (changes requested):

packages/TorneloScoresheet/ios/Podfile
gestureleft: Yep, this is what was causing the linking errors/complaints. I've deployed and tested it and it's all working :)

nocommentcode: Yes but are you sure you've tested from a fresh device with no permissions?
I thought these lines were what made ios ask for camera permission.
Can you just confirm if you start from a fresh device with nothing saved, does it ask you for camera permission and subsequently work fine?

gestureleft: Yep, I tested on the iPad mini, which hadn't had permissions previously (since it hadn't been tested yet), and it successfully asked for permissions. It's lines 16 and 17 that is crucial to the permissions

nocommentcode: oh ok sweet as then

nocommentcode (approved):
lgtm

PR #181, opened 2022-11-07
Fix email game function and errors

nocommentcode:
Fixed the emailing functionality
- Updated env variables to use production instead of staging
- Changed button text upon successfull send
- Update past game modal styling and behaviour on error


scd (approved):
I would have preferred to see two env files, one for production and one for staging?

PR #182, opened 2022-11-07
Board colour options

nocommentcode:
Added new setting for Board Colour Style, which can be either CLASSIC or TORNELO.

### DIFF
--- a/packages/TorneloScoresheet/src/types/GeneralSettingsState.ts
+++ b/packages/TorneloScoresheet/src/types/GeneralSettingsState.ts
@@ -1,12 +1,19 @@
 export enum ChessPieceStyles {
-  TORNELO,
-  CLASSIC,
+  TORNELO = 'TORNELO',
+  CLASSIC = 'CLASSIC',
+}
+
+export enum ChessBoardStyles {
+  TORNELO = 'TORNELO',
+  CLASSIC = 'CLASSIC',
 }
 
 export type GeneralSettings = {
   chessPieceStyle: ChessPieceStyles;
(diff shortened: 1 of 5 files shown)


packages/TorneloScoresheet/src/types/GeneralSettingsState.ts
```diff
@@ -3,10 +3,17 @@ export enum ChessPieceStyles {
   CLASSIC,
 }
 
+export enum ChessBoardStyles {
```
scd, line 6: Are these being saved. They should have explict values if they are being saved in settings. 
Eg 
TORNELO = 'TORNELO',

Otherwise I believe they will be saved in the settings as 0 and 1.

scd (changes requested):
Maybe I have missed something but I think settings with enums should have explicit values, rather then system decide values. Happy for you to recommend otherwise for the context of the scoresheet app.

Also I am not sure of the importance of zindex in your architecture. Having to set this in dropdowns makes me think there is something not ideal? No need to change your architecture for this PR, just putting my comments in.

nocommentcode:
The Z index for the drop down is only because of this strange behaviour when 2 drop down boxes are next to each other:
image


This is linked to the library i'm using for the drop down boxes, perhaps there is a better way to ensure they do not overlap each other but the Z index was what stack overflow said to do to get it like this:
image

scd:
Thanks for letting me know about the zindex limitations. From now on I will assume you are managing this to your library requirements.

scd (approved):

PR #184, opened 2022-11-07
Ux improvements

nocommentcode:
This PR Updates the arbiter navigation menu to match the images provided.
It is missing the go to active game button functionality


packages/TorneloScoresheet/src/components/Toolbar/ArbiterModeNavigation.tsx
```diff
@@ -33,84 +70,135 @@ const ArbiterModeNavigation: React.FC = () => {
...
+    assignGameTransition: (() => void) | undefined | null,
+    viewHistoryTransition: (() => void) | undefined | null,
+    willLoseData: boolean,
+  ): Option[] => {
+    const voidReturn = () => {
+      return;
+    };
+    return [
+      makeNavigationOption(
+        'Select active event',
```
scd, line 84: There is a lot of repeat information and logic in each of these calls. Is it worth considering a wrapper function so you only have to pass in Label and Function. All other parameters are the same and logic the same in each local call to makeNavigationOption.  The reason to do this is that review and ongoing code management will be easier.

Note that you can define a local wrapper function around makeNavigationOption

packages/TorneloScoresheet/src/components/Toolbar/ArbiterModeNavigation.tsx
```diff
@@ -33,84 +70,135 @@ const ArbiterModeNavigation: React.FC = () => {
...
-      ),
-      makeNavigationOption(
-        'Table Pairing',
-        useArbiterRecordingState()?.[1].goBackToTablePairing,
-      ),
-    ],
-    [AppMode.ArbiterTablePairing]: [
-      makeNavigationOption(
-        'Enter Pgn',
+        useArbiterRecordingState()?.[1].goToGameHistory,
```
scd, line 159: Just a general comment about using `[1]` in the code. Without knowing the architecture of the scoresheet app it is difficult to review why you are getting the second variable in a list and what it is?

I am just assuming what you have done here is correct.

Maybe it would be better to use a local variable

```
const [appState,appNav] = useArbiterTablePairingState() ?? []
```

Then it would be far easier to review the logic.

scd (changes requested):

nocommentcode:
@scd This PR is now ready with all the missing functionality,
I've also cleaned up the function to generate the navigation options and am now not using array indexing.

scd (approved):

